PR #4, opened 2021-10-01
IPFS Pool Cache 

snreynolds:
checkpointing the pool cache deploy on ipfs

### DIFF
--- a/bin/app.ts
+++ b/bin/app.ts
@@ -41,6 +41,11 @@ export class RoutingAPIStage extends Stage {
       provisionedConcurrency: number;
       ethGasStationInfoUrl: string;
       chatbotSNSArn?: string;
+      stage: string;
+      route53Arn?: string;
+      pinata_key: string;
+      pinata_secret: string;
+      hosted_zone: string;
     }
   ) {
     super(scope, id, props);
@@ -53,7 +58,12 @@ export class RoutingAPIStage extends Stage {
       nodeRPCPasswordRinkeby,
       provisionedConcurrency,
       ethGasStationInfoUrl,
-      chatbotSNSArn
+      chatbotSNSArn,
+      stage, 
+      route53Arn,
+      pinata_key,
+      pinata_secret,
+      hosted_zone,
     } = props;
 
     const { url } = new RoutingAPIStack(this, 'RoutingAPI', {
@@ -66,6 +76,11 @@ export class RoutingAPIStage extends Stage {
       provisionedConcurrency,
       ethGasStationInfoUrl,
       chatbotSNSArn,
+      stage,
+      route53Arn,
+      pinata_key,
+      pinata_secret,
+      hosted_zone,
     });
     this.url = url;
   }
@@ -131,6 +146,8 @@ export class RoutingAPIPipeline extends Stack {
       }
     );
 
+    // grab secrets for route53, pinata keys, hosted zone
+
     // Beta us-east-2
     const betaUsEast2Stage = new RoutingAPIStage(this, 'beta-us-east-2', {
       env: { account: '145079444317', region: 'us-east-2' },
@@ -150,7 +167,12 @@ export class RoutingAPIPipeline extends Stack {
         .toString(),
       provisionedConcurrency: 20,
       ethGasStationInfoUrl: ethGasStationInfoUrl.secretValue.toString(),
-      chatbotSNSArn: 'arn:aws:sns:us-east-2:644039819003:SlackChatbotTopic'
+      chatbotSNSArn: 'arn:aws:sns:us-east-2:644039819003:SlackChatbotTopic',
+      stage: 'beta',
+      route53Arn: 'from secrets manager',
+      pinata_key: 'from secrets manager',//from secrets manager,
+      pinata_secret: 'from secrets manager',// from secrets manager,
+      hosted_zone:  'from secrets manager,'// from secrets manager, just trying to get it to compile locally
     });
 
     const betaUsEast2AppStage = pipeline.addApplicationStage(betaUsEast2Stage);
@@ -181,7 +203,12 @@ export class RoutingAPIPipeline extends Stack {
         .toString(),
       provisionedConcurrency: 100,
       ethGasStationInfoUrl: ethGasStationInfoUrl.secretValue.toString(),
-      chatbotSNSArn: 'arn:aws:sns:us-east-2:644039819003:SlackChatbotTopic'
+      chatbotSNSArn: 'arn:aws:sns:us-east-2:644039819003:SlackChatbotTopic',
+      stage: 'prod',
+      route53Arn: 'from secrets manager',
+      pinata_key: 'from secrets manager',//from secrets manager,
+      pinata_secret: 'from secrets manager',// from secrets manager,
+      hosted_zone:  'from secrets manager,'// from secrets manager, just trying to get it to compile locally
     });
 
     const prodUsEast2AppStage = pipeline.addApplicationStage(prodUsEast2Stage);
@@ -249,6 +276,11 @@ new RoutingAPIStack(app, 'RoutingAPIStack', {
   throttlingOverride: process.env.THROTTLE_PER_FIVE_MINS,
   ethGasStationInfoUrl: process.env.ETH_GAS_STATION_INFO_URL!,
   chatbotSNSArn: process.env.CHATBOT_SNS_ARN,
+  stage: process.env.STAGE!,
+  route53Arn: process.env.ROLE_ARN,
+  pinata_key: process.env.PINATA_API_KEY!,
+  pinata_secret: process.env.PINATA_API_SECRET!,
+  hosted_zone: process.env.HOSTED_ZONE!
 });
 
 new RoutingAPIPipeline(app, 'RoutingAPIPipelineStack', {
--- a/bin/stacks/routing-caching-stack.ts
+++ b/bin/stacks/routing-caching-stack.ts
@@ -1,18 +1,22 @@
 import * as aws_events from '@aws-cdk/aws-events';
 import * as aws_events_targets from '@aws-cdk/aws-events-targets';
 import * as aws_iam from '@aws-cdk/aws-iam';
-import * as aws_cloudwatch from '@aws-cdk/aws-cloudwatch';
-import * as aws_cloudwatch_actions from '@aws-cdk/aws-cloudwatch-actions';
 import * as aws_lambda from '@aws-cdk/aws-lambda';
 import * as aws_lambda_nodejs from '@aws-cdk/aws-lambda-nodejs';
 import * as aws_s3 from '@aws-cdk/aws-s3';
-import * as aws_sns from '@aws-cdk/aws-sns';
 import * as cdk from '@aws-cdk/core';
 import { Construct, Duration } from '@aws-cdk/core';
 import * as path from 'path';
+import dotenv from 'dotenv';
+import { PolicyStatement } from '@aws-cdk/aws-iam';
+dotenv.config()
 
 export interface RoutingCachingStackProps extends cdk.NestedStackProps {
-  chatbotSNSArn?: string;
+  stage: string;
+  route53Arn?: string;
+  pinata_key: string;
+  pinata_secret: string;
+  hosted_zone: string;
 }
 export class RoutingCachingStack extends cdk.NestedStack {
   public readonly poolCacheBucket: aws_s3.Bucket;
@@ -22,11 +26,17 @@ export class RoutingCachingStack extends cdk.NestedStack {
   constructor(scope: Construct, name: string, props: RoutingCachingStackProps) {
     super(scope, name, props);
 
-    const { chatbotSNSArn } = props;
-
     this.poolCacheBucket = new aws_s3.Bucket(this, 'PoolCacheBucket');
     this.poolCacheKey = 'poolCache.json';
 
+    const {
+      stage,
+      route53Arn,
+      pinata_key,
+      pinata_secret,
+      hosted_zone
+    } = props;
+  
     const lambdaRole = new aws_iam.Role(this, 'RoutingLambdaRole', {
       assumedBy: new aws_iam.ServicePrincipal('lambda.amazonaws.com'),
       managedPolicies: [
@@ -36,6 +46,11 @@ export class RoutingCachingStack extends cdk.NestedStack {
         aws_iam.ManagedPolicy.fromAwsManagedPolicyName('CloudWatchFullAccess'),
       ],
     });
+
+    lambdaRole.addToPolicy(new PolicyStatement({
+      resources: ['*'], 
+      actions: ['sts:AssumeRole']
+    }))
 
     const region = cdk.Stack.of(this).region;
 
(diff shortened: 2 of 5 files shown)


bin/app.ts
```diff
@@ -249,6 +276,11 @@ new RoutingAPIStack(app, 'RoutingAPIStack', {
   throttlingOverride: process.env.THROTTLE_PER_FIVE_MINS,
   ethGasStationInfoUrl: process.env.ETH_GAS_STATION_INFO_URL!,
   chatbotSNSArn: process.env.CHATBOT_SNS_ARN,
+  stage: process.env.STAGE!,
```
willpote, line 279: Nit: We can just hardcode this to 'local' when we merge

bin/stacks/routing-caching-stack.ts
```diff
@@ -37,6 +53,11 @@ export class RoutingCachingStack extends cdk.NestedStack {
       ],
     });
 
+    lambdaRole.addToPolicy(new PolicyStatement({
+      resources: ['*'],
```
willpote, line 57: Best to always scope down policies to have minimum positions. You can specify the role arn we're assuming here

bin/app.ts
```diff
@@ -41,6 +41,11 @@ export class RoutingAPIStage extends Stage {
       provisionedConcurrency: number;
       ethGasStationInfoUrl: string;
       chatbotSNSArn?: string;
+      stage: string;
+      route53Arn?: string;
+      pinata_key: string;
+      pinata_secret: string;
+      hosted_zone: string;
```
willpote, line 48: Can we make these optional, so that we can deploy the API locally without needing to specify these?

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,112 @@
...
+  var route53 = new Route53();
+  sts.assumeRole(stsParams, (err: any, data: any) => {
+    if (err) {
+      log.error({ err }, `Error assuming role`);
+    } else {
+      credentials = data.Credentials;
+      route53 = new Route53({ credentials: credentials });
+      log.info(`Role assumed`);
+    }
+  });
```
willpote, line 48: We can await/async this:

`await sts.assumeRole(stsParams).promise()`

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,112 @@
...
+    const subgraphProvider = new SubgraphProvider(chains[i], 3, 15000);
+    const pools = await subgraphProvider.getPools();
+    // TODO : filter pools 
+    const poolString = JSON.stringify(pools);
+
+    let fileName = names[i]
+
+    // create directory and file
+    fs.mkdirSync(DIRECTORY, { recursive: true });
+    fs.writeFileSync(DIRECTORY.concat(fileName), poolString);
```
willpote, line 60: Nit: best practice to use Node's `path` module for all things directory/path related as it handles awkward things that cause bugs like extra slashes and works cross platform. 

`path.join` would be good here: https://nodejs.org/api/path.html#path_path_join_paths

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,112 @@
...
+      result = await pinata.pinFromFS(DIRECTORY);
+      let url = `https://ipfs.io/ipfs/${result.IpfsHash}`;
+      results.push(result.IpfsHash)
+      
+      log.info(
+        `Succcessful pinning. IPFS hash: ${result.IpfsHash} and url : ${url}`
+      );
+    } catch (err) {
+        log.error({ err }, 'Error pinning');
+      }
```
willpote, line 77: If we fail to pin, we would want to exit immediately and not try and update the DNS. Should throw `err` here

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,112 @@
...
+      
+      // link all chains pool data to DNS
+      for (let i = 0; i < chains.length; i++) {
+        const result = results[i]
+        const fileName = names[i]
+        var params = {
+            ChangeBatch: {
+              Changes: [
+                {
+                  Action: 'CREATE',
```
willpote, line 87: Think we want to do an `UPSERT` here

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,112 @@
...
+      );
+    } catch (err) {
+        log.error({ err }, 'Error pinning');
+      }
+      
+      // link all chains pool data to DNS
+      for (let i = 0; i < chains.length; i++) {
+        const result = results[i]
+        const fileName = names[i]
+        var params = {
```
willpote, line 83: I'm a bit confused why we are doing this in a loop? 

I think we just need to set 
`dnslink=/ipfs/${result}` 
once and thats it

That would link beta.api.uniswap.org to the root IPFS directory. Then when we do `ipfs.io/ipns/beta.api.uniswap.org/v1/pools/mainnet.json` DNSlink would resolve us to:
`ipfs.io/ipfs/<CID>/pools/mainnet`

The examples here look similar to what we want: https://dnslink.io/ , looks like they just linked the base CID on dnslink, then they can browse all the files in the directory from the ipns link

lib/cron/cache-pools-ipfs.ts
willpote: Also indenting looks slightly off

bin/stacks/routing-caching-stack.ts
```diff
@@ -70,9 +91,41 @@ export class RoutingCachingStack extends cdk.NestedStack {
 
     this.poolCacheBucket.grantReadWrite(poolCachingLambda);
 
+    const ipfsPoolCachingLambda = new aws_lambda_nodejs.NodejsFunction(
```
willpote, line 94: Fine for testing but when we merge we should only really create this resource if `stage == beta` (and eventually prod). It would just fail locally since we don't have permissions to access the route53 role anyway

bin/stacks/routing-caching-stack.ts
```diff
@@ -70,9 +91,41 @@ export class RoutingCachingStack extends cdk.NestedStack {
...
+      }
+    );
+
     new aws_events.Rule(this, 'SchedulePoolCache', {
       schedule: aws_events.Schedule.rate(Duration.minutes(2)),
-      targets: [new aws_events_targets.LambdaFunction(poolCachingLambda)],
+      targets: [new aws_events_targets.LambdaFunction(poolCachingLambda), new aws_events_targets.LambdaFunction(ipfsPoolCachingLambda)],
     });
 
     const lambdaAlarmErrorRate = new aws_cloudwatch.Alarm(this, 'RoutingAPI-PoolCacheToS3LambdaError', {
```
willpote, line 131: Would be great to add an alarm for the new lambda (also only created if on beta)

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,112 @@
...
+import fs from 'fs';
+
+// if (process.env.STAGE == 'beta') {}
+
+const DIRECTORY = '/tmp/pools/';
+
+// add more chains here
+const chains = [ChainId.MAINNET, ChainId.RINKEBY]
+const names = ["mainnet.txt", "rinkeby.txt"]
+
```
willpote, line 15: Nit: prefer a complex obj

`const chains: { fileName: string, chain: ChainId }[] = ..`

and later

`const { fileName, chain } = chains[i];`

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,112 @@
+import pinataSDK from '@pinata/sdk';
+import { ChainId, SubgraphProvider } from '@uniswap/smart-order-router';
+import { EventBridgeEvent, ScheduledHandler } from 'aws-lambda';
+import { Route53, STS } from 'aws-sdk';
+import { default as bunyan, default as Logger } from 'bunyan';
+import fs from 'fs';
+
+// if (process.env.STAGE == 'beta') {}
+
+const DIRECTORY = '/tmp/pools/';
```
willpote, line 10: Can we get the v1 in here? `/tmp/v1/pools`? then we pin `v1` directory?

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,112 @@
...
+
+// add more chains here
+const chains = [ChainId.MAINNET, ChainId.RINKEBY]
+const names = ["mainnet.txt", "rinkeby.txt"]
```
willpote, line 14: Can we remove the `.txt`? ideally the ipns is simply like `beta.api.uniswap.org/v1/pools/mainnet`

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+  });
+
+  for (let i = 0; i < chains.length; i ++ ) {
+    const subgraphProvider = new SubgraphProvider(chains[i], 3, 15000);
+    const pools = await subgraphProvider.getPools();
+    // TODO : filter pools 
+    const poolString = JSON.stringify(pools);
+
+    let fileName = names[i]
```
willpote, line 56: const

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,112 @@
...
+    fs.writeFileSync(DIRECTORY.concat(fileName), poolString);
+  }
+
+    // pins everything under '/tmp/` which should include mainnet.txt and rinkeby.txt
+    // only have to pin once for all chains
+    let result;
+    const results : string[] = []
+    try {
+      result = await pinata.pinFromFS(DIRECTORY);
+      let url = `https://ipfs.io/ipfs/${result.IpfsHash}`;
```
willpote, line 69: const

lib/cron/cache-pools-ipfs.ts
willpote: Oh also we need a way to specify v3 vs v2 protocol.. `pools/mainnet/v3`?

lib/cron/cache-pools-ipfs.ts
snreynolds: Ah I see. You're right. We don't need to loop if we attach the DNS to the parent directory. Good catch - missed this!

lib/cron/cache-pools-ipfs.ts
snreynolds: Do you mean v3? Why v1?

lib/cron/cache-pools-ipfs.ts
snreynolds: Hm, so I can't remove .txt because I need to write to a file otherwise I get an error about writing to a directory. Can be a different extension though

lib/cron/cache-pools-ipfs.ts
willpote: Oh yeah sorry this was a confusing comment. Our API is versioned, so the current quote endpoint is api.uniswap.org/v1/quote where v1 means version 1 of the API. 

So to fit `pools` endpoint into that structure we would want to create a folder 'v1' on ipfs. Then our final IPNS urls would be like: `ipfs.io/ipns/api.uniswap.org/v1/pools/mainnet/v3`

lib/cron/cache-pools-ipfs.ts
willpote: Hmm ok I think .json makes more sense then

lib/cron/cache-pools-ipfs.ts
snreynolds: oh I see. The only issue is that the there needs to be a file that we write to. So we could do something like:
`ipfs.io/ipns/api.uniswap.org/v1/pools/v3/mainnet.txt`. Or open to other suggestions, just need a .txt or .json

lib/cron/cache-pools-ipfs.ts
snreynolds: Ok sounds good - updated with `ipfs.io/ipns/api.uniswap.org/v1/pools/v3/mainnet.json` but lmk if we should change it

bin/app.ts
```diff
@@ -150,7 +190,12 @@ export class RoutingAPIPipeline extends Stack {
         .toString(),
       provisionedConcurrency: 20,
       ethGasStationInfoUrl: ethGasStationInfoUrl.secretValue.toString(),
-      chatbotSNSArn: 'arn:aws:sns:us-east-2:644039819003:SlackChatbotTopic'
+      chatbotSNSArn: 'arn:aws:sns:us-east-2:644039819003:SlackChatbotTopic',
+      stage: 'beta',
```
willpote, line 194: Can you make a `Stage` enum?

bin/stacks/routing-caching-stack.ts
```diff
@@ -37,6 +47,13 @@ export class RoutingCachingStack extends cdk.NestedStack {
       ],
     });
 
+    lambdaRole.addToPolicy(
+      new PolicyStatement({
+        resources: [route53Arn!],
```
willpote, line 52: Does it still deploy if route53Arn is undefined? I think IAM does some basic validation of policies during deployment so I would expect the deployment to fail?

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    const sts = new STS();
+    const stsParams = {
+      RoleArn: process.env.ROLE_ARN!,
+      RoleSessionName: `UpdateApiRoute53Role`,
+    };
+
+    // init route53 with credentials
+    let data;
+    var route53 = new Route53();
```
willpote, line 42: const

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,118 @@
...
+        ],
+      },
+      HostedZoneId: process.env.HOSTED_ZONE!,
+    };
+    try {
+      const data = await route53.changeResourceRecordSets(params).promise();
+      log.info(`Successful record update: ${data}`);
+    } catch (err) {
+      log.error({ err }, 'Error updating DNS');
+    }
```
willpote, line 114: Throw error

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,118 @@
...
+    // pins everything under '/tmp/` which should include mainnet.txt and rinkeby.txt
+    // only have to pin once for all chains
+    let result;
+    let hash;
+    try {
+      result = await pinata.pinFromFS(PARENT);
+      const url = `https://ipfs.io/ipfs/${result.IpfsHash}`;
+      hash = result.IpfsHash;
+
+      log.info(`Succcessful pinning. IPFS hash: ${hash} and url : ${url}`);
```
willpote, line 83: Nit:  good to log the whole result object too

`log.info({ result }, 'Successful...')

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,118 @@
...
+      const secretAccess = data?.Credentials?.SecretAccessKey;
+      route53 = new Route53({
+        credentials: {
+          accessKeyId: accessKeyId!,
+          secretAccessKey: secretAccess!,
+        },
+      });
+    } catch (err: any) {
+      log.error({ err }, 'Route53 not initialized with correct credentials');
+    }
```
willpote, line 61: We want to throw the error so that the lambda fails, and the alarm will get triggered

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,118 @@
+import pinataSDK from '@pinata/sdk';
+import { ChainId, SubgraphProvider } from '@uniswap/smart-order-router';
+import { EventBridgeEvent, ScheduledHandler } from 'aws-lambda';
+import { Route53, STS } from 'aws-sdk';
+import { default as bunyan, default as Logger } from 'bunyan';
+import fs from 'fs';
+
+if (process.env.STAGE == 'beta') {
```
willpote, line 8: You shouldn't really need this since you only create the lambda in Cloudformation if we are on beta but I get it if you added just for insurance. 

In any case, I think having the if statement surrounding everything like this would cause issues if we had unit tests as the module locally. 

I'd just put this inside the handler and do an early return `if (process.env.STAGE != Stage.BETA) { //log something; return; }`

lib/cron/cache-pools-ipfs.ts
```diff
@@ -0,0 +1,118 @@
...
+
+    // init route53 with credentials
+    let data;
+    var route53 = new Route53();
+    try {
+      data = await sts.assumeRole(stsParams).promise();
+    } catch (err) {
+      log.error({ err }, `Error assuming role`);
+    }
+
```
willpote, line 48: Throw error

bin/stacks/routing-caching-stack.ts
snreynolds: Ok will do!

snreynolds: Yeah, we'd need to require the route53 arn so I just wrapped it in an if statement

bin/stacks/routing-caching-stack.ts
```diff
@@ -1,17 +1,26 @@
...
 import * as aws_lambda from '@aws-cdk/aws-lambda';
 import * as aws_lambda_nodejs from '@aws-cdk/aws-lambda-nodejs';
 import * as aws_s3 from '@aws-cdk/aws-s3';
 import * as aws_sns from '@aws-cdk/aws-sns';
 import * as cdk from '@aws-cdk/core';
 import { Construct, Duration } from '@aws-cdk/core';
+import dotenv from 'dotenv';
 import * as path from 'path';
+import { STAGE } from '../app';
+dotenv.config();
```
willpote, line 16: Shouldn't need this

willpote (approved):
Looks good!

PR #8, opened 2021-10-08
Add swapToRatio endpoint

ewilz:
This PR includes some npm package updates to match smart-order-router sdk versions. I'm thinking to push those first in their own PR to separate out those updates from this work.

Todo:

- [x] DRY
- [x] Separate out sdk package updates into separate PR


jest.config.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
     '^.+\\.tsx?$': 'ts-jest',
   },
   testEnvironment: 'node',
-  testTimeout: 15000,
+  testTimeout: 30000,
```
ewilz, line 8: 😬  lol this has to do with my high volume 500 million dollar integration test trade for swap + add. i can limit the tests to 100-200 million if preferred.

lib/handlers/shared.ts
```diff
@@ -0,0 +1,66 @@
+import { AlphaRouterConfig } from '@uniswap/smart-order-router';
```
ewilz, line 1: mmm open to ideas on where to put these things

lib/handlers/injector-sor.ts
```diff
@@ -0,0 +1,171 @@
...
+  quoteProvider: QuoteProvider;
+};
+
+export interface ContainerInjected {
+  dependencies: {
+    [chainId in ChainId]?: ContainerDependencies;
+  }
+}
+
+export abstract class InjectorSOR<R, P> extends Injector<
```
ewilz, line 63: quote & quoteToRatio differ in Router and Params. I can extend these letters to signify these two things.

test/integ/quote-to-ratio.test.ts
```diff
@@ -0,0 +1,357 @@
...
+
+  return token;
+
+}
+
+function parseFraction(fraction: ResponseFraction): Fraction {
+  return new Fraction(JSBI.BigInt(fraction.numerator), JSBI.BigInt(fraction.denominator))
+}
+
+describe('quote-to-ratio', () => {
```
willpote, line 66: Can we get some 4xx failure case tests?

Good to test some of the basic validation cases just to check that validation is hooked up (e.g. incorrect address returns 4xx etc.), as well as cases where we don't find a swap? (maybe we can set maxIterations to 0 or 1? or maybe passing in token0Balance = 0 and token1Balance = 0?)

Also what happens if I want to do a range order, but I already have 100% of the token required and 0 of the token not required? Do we still recommend a swap? Do we return 200 or 4xx in that case?

lib/handlers/quote-to-ratio/quote-to-ratio.ts
```diff
@@ -0,0 +1,422 @@
...
+      token1Balance,
+      position,
+      {
+        errorTolerance: errorToleranceFraction,
+        maxIterations
+      },
+      swapParams,
+      routingConfig,
+    );
+    log.info('made it')
```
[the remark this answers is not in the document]
ewilz: haha woops I have a lot of leftover logs I forgot about.

lib/handlers/quote-to-ratio/quote-to-ratio.ts
```diff
@@ -0,0 +1,422 @@
...
+      log.error(
+        `Could not find pool.`
+      );
+      return { statusCode: 400, errorCode: 'Could not find pool.' };
+    }
+    const position = new Position({
+      pool,
+      tickLower,
+      tickUpper,
+      liquidity: 1,
```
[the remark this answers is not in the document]
ewilz: hmm...it's weird because liquidity plays absolutely no role in any calculation. we define a position to get the bounds, in many cases this will be a hypothetical position with no liquidity yet.

lib/handlers/quote-to-ratio/quote-to-ratio.ts
```diff
@@ -0,0 +1,422 @@
...
+    let newRatioDecimal: string
+    if (zeroForOne) {
+      token0BalanceUpdated = token0Balance.subtract(trade.inputAmount)
+      token1BalanceUpdated = token1Balance.add(trade.outputAmount)
+      optimalRatioAdjusted = optimalRatio
+      optimalRatioDecimal = optimalRatioAdjusted.toFixed(token0.wrapped.decimals)
+      newRatioDecimal = new Fraction(
+        token0BalanceUpdated.quotient.toString(),
+        token1BalanceUpdated.quotient.toString()
+      ).toFixed(token0.wrapped.decimals)
```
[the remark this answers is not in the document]
ewilz: since this is in terms of x token0 for token1, I do the decimal in terms of token0. Since the numbers on chain tend to be multiplied by e18, I thought it might be useful to keep the precision relatively big

PR #15, opened 2021-11-03
prettier linter

ewilz:
When I run the linter this is the results I get. If this matches up with what you get @willpote or @snreynolds, maybe we want to merge this. Not sure if I'm all configured right.

also, I'd maybe want to add a hook here to run the linter automatically after a commit and write if updates are needed, similar to the other repos.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,33 @@
+name: Lint
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  run-linters:
+    name: Run linters
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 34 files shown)


willpote:
Thanks for doing this!

So what we use in the interface/sdk is:

```
"prettier": {
    "printWidth": 120,
    "semi": false,
    "singleQuote": true
  }
```

so we should probably use this config. Rip semi colons

willpote:
Also yes if you can set up the hook that would be great! Have been meaning to do it but haven't had time

ewilz:
RIP git blame. 

ready for  review @willpote

willpote (approved):
Cheers!

PR #17, opened 2021-11-05
SwapToRatio: ratio fullfilled + token order errors

ewilz:


lib/handlers/quote-to-ratio/quote-to-ratio.ts
```diff
@@ -100,6 +97,14 @@ export class QuoteToRatioHandler extends APIGLambdaHandler<
       }
     }
 
+    if (token0.wrapped.address > token1.wrapped.address) {
```
ewilz, line 100: I think it would be too confusing to hand the tokens in one order, and get the results back for the different ordering, so thought throwing an error instead would prevent potential headaches.

ewilz: then in the tests below, had to subsequently make sure everything was in the right order ^_^

willpote (approved):
Looks good, though need to handle merge conflicts now we moved to mocha

PR #21, opened 2021-11-17
V2 pools

snreynolds:


lib/cron/cache-pools-ipfs.ts
```diff
@@ -60,13 +65,24 @@ const handler: ScheduledHandler = async (event: EventBridgeEvent<string, void>)
...
-    // create directory and file
-    fs.mkdirSync(DIRECTORY, { recursive: true })
-    fs.writeFileSync(DIRECTORY.concat(fileName), poolString)
+    // create directory and file for v3
+    const directoryV3 = DIRECTORY.concat(VERSION.V3).concat('/')
+    fs.mkdirSync(directoryV3, { recursive: true })
+    fs.writeFileSync(directoryV3.concat(fileName), poolString)
+
+    // create directory and file for v2
+    const subgraphProviderV2 = new V2SubgraphProvider(chain, 3, 15000)
```
willpote, line 78: Getting stuff from the V2 Subgraph is actually way slower, it won't finish in 15s. I'd just remove the 15000 and use the default (which is around 5 mins)

willpote (approved):

PR #25, opened 2021-11-19
Hardhat tests

willpote:
Jest -> Mocha in order to make hardhat tests work.

Simulate swaps using hardhat fork for the quote integ tests


hardhat.config.js
```diff
@@ -0,0 +1,16 @@
+require('@nomiclabs/hardhat-ethers')
+require('dotenv').config()
+
+module.exports = {
+  defaultNetwork: 'hardhat',
+  networks: {
+    hardhat: {
+      blockGasLimit: 150_000_000,
```
ewilz, line 8: this is bc 10x block gas limit for??

hardhat.config.js
```diff
@@ -0,0 +1,16 @@
...
+      forking: {
+        enabled: true,
+        url: `${process.env.ARCHIVE_NODE_RPC}`,
+        blockNumber: 12972642,
```
ewilz, line 12: hmm...is this for predictable test results?? wonder if there'd be a way to test things without needing exact numbers...haven't thought it through yet though

hardhat.config.js
willpote: Actually this isn't needed since we don't do the static calls in the tests. This was copy pasted from another project where I was doing huge static calls. Will remove

hardhat.config.js
willpote: Actually this was copy pasted over too and isn't needed. 

The tests already don't require exact numbers. Currently I check the users balance before and after the swap

e.g. For example for an exactIn swap of 100, i check the user has 100 less of tokenIn after the swap. I don't check tokenOut, but thinking more about it- i think it would be valuable to check token out is within slippage percentage of the quote. Ill add some more checks

test/integ/quote.test.ts
```diff
@@ -1,808 +1,1078 @@
...
+    : tokensSwapped.subtract(tokensQuoted)
+  const percentDiff = tokensDiff.asFraction.divide(tokensQuoted.asFraction)
+  /* console.log(
+    `5pc: ${new Fraction(5, 100).toFixed(10)} tokens diff: ${tokensDiff.toExact()} percent diff: ${percentDiff.toFixed(
+      5
+    )} ${percentDiff.quotient} ${percentDiff.remainder.toFixed(5)} (${percentDiff.numerator}/${
+      percentDiff.denominator
+    }), quote: ${tokensQuoted.toExact()}, rceived: ${tokensSwapped.toExact()}`
+  ) */
+  expect(percentDiff.lessThan(new Fraction(5, 100))).to.be.true
```
ewilz, line 65: where does this figure come from??

willpote: All the tests specify 5% for slippage. Refactored this to a constant

PR #43, opened 2022-05-16
fix gorli integ tests

snreynolds:


willpote (approved):
Approving, you can add Goerli to the pool cache lambdas though

PR #58, opened 2022-07-13
Fixed batch params override issue cause celo quotes to fail

raghava-pamula:
Fixes issue where batch params for Celo in SOR get overridden by default batch params in routing-api

### DIFF
--- a/lib/handlers/injector-sor.ts
+++ b/lib/handlers/injector-sor.ts
@@ -303,7 +303,7 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
     )
 
     for (const { chainId, dependencies } of dependenciesByChainArray) {
-      dependenciesByChain[chainId] = dependencies as ContainerDependencies
+      dependenciesByChain[chainId] = dependencies
     }
 
     return {
--- a/lib/handlers/injector-sor.ts
+++ b/lib/handlers/injector-sor.ts
@@ -157,7 +157,7 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
 
         // Some providers like Infura set a gas limit per call of 10x block gas which is approx 150m
         // 200*725k < 150m
-        let quoteProvider: V3QuoteProvider|undefined = undefined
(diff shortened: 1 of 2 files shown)


lib/handlers/injector-sor.ts
```diff
@@ -324,7 +305,9 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
     )
 
     for (const { chainId, dependencies } of dependenciesByChainArray) {
-      dependenciesByChain[chainId] = dependencies
+      if (dependencies) {
```
zhongeric, line 308: is this `if` statement needed?

lib/handlers/quote/injector.ts
```diff
@@ -91,6 +91,9 @@ export class QuoteHandlerInjector extends InjectorSOR<
     let router
     switch (algorithm) {
       case 'legacy':
+        if (!v3QuoteProvider) {
+          throw new Error('V3 Quote Provider not set for Legacy Routings')
+        }
```
zhongeric, line 96: lgtm, only thing is im not sure if this is the behavior we want for legacy router, because essentially it would break legacy router for all chains except those specifically defined here (all except Optimism and Arb)

raghava-pamula: what do you suggest instead

raghava-pamula: I think if the v3QuoteProvider is not set here, Legacy Router would break anyway since it doesn't have a fallback quote provider right?

zhongeric: Yeah it doesn't have a fallback and expects it to be passed into the constructor. 

If the quoteProvider is undefined, we could create the quoteProvider here with the default settings and just pass into legacy router only, which would mimic the previous behavior. Since I assume we would want legacy to still work

willpote: Agree with Eric, lets put the old behavior here for legacy

zhongeric (approved):

PR #62, opened 2022-07-18
Tenderly simulations

raghava-pamula:
### Feature Added: Simulation Via Tenderly
Quote API now takes optional param `?simulate=true`. If true, we create a fork from the current block head and:
1. Route as usual
1. Simulate approval to the tokenInAddress
2. Simulate the transaction using the calldata from SwapRoute
We simulate these transactions from the recipients address, since in most cases, this is the same as the from address, which should have the necessary balance already for the swap, saving us the need to perform state override for the token contract or simulate transaction minting the ERC20 to a dummy address.

The response now looks like [10 USDC -> UNI Exact Input]:
`{'methodParameters': {'calldata': '0x5ae401dc0000000000000000000000000000000000000000000000000000000062d4e7af00000000000000000000000000000000000000000000000000000000000000400000000000000000000000000000000000000000000000000000000000000001000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000e404e45aaf000000000000000000000000a0b86991c6218b36c1d19d4a2e9eb0ce3606eb480000000000000000000000001f9840a85d5af5bf1d1762f925bdaddc4201f98400000000000000000000000000000000000000000000000000000000000001f400000000000000000000000063946551716781c32f0269f87dc08521818b6292000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000011b02d0d516000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000', 'value': '0x00'}, 'blockNumber': '15164553', 'amount': '10', 'amountDecimals': '0.00001', 'quote': '1276299134500', 'quoteDecimals': '0.0000012762991345', 'quoteGasAdjusted': '-624894577346256745', 'quoteGasAdjustedDecimals': '-0.624894577346256745', 'gasUseEstimateQuote': '624895853645391245', 'gasUseEstimateQuoteDecimals': '0.624895853645391245', 'gasUseEstimate': '113000', 'gasUseEstimateUSD': '4.554461', 'gasPriceWei': '28692499874', 'route': [[{'type': 'v3-pool', 'address': '0xaB2044f105c43C25b1de3Ee27504f0B889CE5953', 'tokenIn': {'chainId': 1, 'decimals': '6', 'address': '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48', 'symbol': 'USDC'}, 'tokenOut': {'chainId': 1, 'decimals': '18', 'address': '0x1f9840a85d5aF5bf1D1762F925BDADdC4201F984', 'symbol': 'UNI'}, 'fee': '500', 'liquidity': '109371341663368', 'sqrtRatioX96': '210389758143247563331050', 'tickCurrent': '-256791', 'amountIn': '10', 'amountOut': '1276299134500'}]], 'routeString': '[V3] 100.00% = USDC -- 0.05% [0xaB2044f105c43C25b1de3Ee27504f0B889CE5953] --> UNI', 'quoteId': '76275', 'simulatedTxReceipt': '0x01fc13'}`

The only new field is 'simulatedTxReceipt': '0x01fc13'. The value is a hexstring representing the units of gas in wei used to land the transaction on chain in the tenderly fork. We can choose to send back more information from the TransactionReceipt, but only gas estimation is needed currently.

### Benchmarking
[Compiled data on benchmarking+accuracy comparison of gas estimate vs submitting transaction on chain](https://docs.google.com/spreadsheets/d/1TyX--YZ9hFHr4YhtmWG-vX_mWJVflW5sh0k3pK3VGL8/edit#gid=95141336)
Notably:

- Latency increases from 1.74 seconds to 3.48 seconds when simulating (N=200)
- The tenderly gasUsed isn't that accurate. I noticed that the callData that I get from querying routing-api locally is different than the callData for the same transaction with the same pool that lands on chain when I submit a transaction through the UI. 
- I went into tenderly dashboard and used the same callData and blockNumber as the transaction that landed on chain, and the estimate was much much better so I think the issue is just that I'm not simulating with the same calldata as what I'm getting using the official uniswap interface from the production routing-api. Not sure why my local routing-api is returning different calldata, and notably it is much more compact than the prod api response.

### TODO
- [IN PROGRESS] - Add specs/integration tests
- [DONE] - Investigate why calldata is different locally versus in production. If this is solved, then the tenderly simulation gas estimates should be very close to reality.
- [DONE - No more forks ] - Add management for Tenderly Forks. We can cache forks and also delete them when they're 10 blocks out of date. We can manage up to K forks per blockNumber and share them between simulations so as to rate limit and also not create too many forks. The latency in sharing the resource should be lower than creating a new fork every single time. This all depends on what we're able to negotiate with Tenderly too.
- [NOT NEEDED] - Instead of simulating the approval, we can try to set a state override on the tokenInAddress. This should reduce latency, and the tradeoff is the code will be more complicated.


lib/handlers/schema.ts
```diff
@@ -37,27 +37,30 @@ export type V2PoolInRoute = {
...
+    gasUseEstimateUSD: Joi.string().required(),
+    gasPriceWei: Joi.string().required(),
+    blockNumber: Joi.string().required(),
+    route: Joi.array().items(Joi.any()).required(),
+    routeString: Joi.string().required(),
+    methodParameters: Joi.object({
+      calldata: Joi.string().required(),
+      value: Joi.string().required(),
+    }).optional(),
+    simulatedTxReceipt: Joi.string().optional(),
```
willpote, line 61: We should just override the regular gas fields, in addition to the gas used, we would also want to return an accurate gasUseEstimateUSD, gasUseEstimateQuote, etc etc

lib/handlers/tenderly-sim/tenderly-sim.ts
```diff
@@ -0,0 +1,122 @@
...
+export interface ISimulator {
+  simulateTx: (
+    chainId: number,
+    hexData: string,
+    tokenInAddress: string,
+    fromAddress: string,
+    blockNumber: number
+  ) => Promise<TransactionReceipt>
+}
+export class TenderlyProvider implements ISimulator {
```
willpote, line 31: We should move all this into SOR and make it a regular part of the business logic there. 

Simulate=true/false should be a flag in the `SwapOptions` type, and if its true we should override all the gas related fields that are returned to be the more accurate ones from the simulation.

This is a very useful feature that we should also expose to people that are using the SOR directly, and potentially to things like the widget that use the SOR client side.

lib/handlers/tenderly-sim/tenderly-sim.ts
```diff
@@ -0,0 +1,122 @@
...
+export const TENDERLY_FORK_API_URL = (FORK_ID: string): string => `https://rpc.tenderly.co/fork/${FORK_ID}`
+
+// API For POST
+export const POST_TENDERLY_FORK_API_URL = (
+  TENDERLY_BASE_URL: string,
+  TENDERLY_USER: string,
+  TENDERLY_PROJECT: string
+) => `${TENDERLY_BASE_URL}/api/v1/account/${TENDERLY_USER}/project/${TENDERLY_PROJECT}/fork`
+
+export const APPROVE_TOKEN_FOR_TRANSFER =
```
willpote, line 19: Whats this?

lib/handlers/tenderly-sim/tenderly-sim.ts
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  private async getFork(params: { chainId: number; blockNumber: number }): Promise<JsonRpcProvider> {
+    log.info({ params: params }, 'Getting Tenderly Fork')
+    // Assume Fork of the block does not yet exist in our tenderly project
+    // TODO implement storage and management of tenderly forks
+    const fork = await this.createFork(params)
+    return fork
+  }
+  private createFork(params: { chainId: number; blockNumber: number }): Promise<JsonRpcProvider> {
```
willpote, line 100: Spacing above

lib/handlers/tenderly-sim/tenderly-sim.ts
```diff
@@ -0,0 +1,122 @@
...
+      save_if_fails: true,
+    }
+    return axios
+      .post(POST_TENDERLY_FORK_API_URL(this.TENDERLY_BASE_URL, this.TENDERLY_USER, this.TENDERLY_PROJECT), body, opts)
+      .then(
+        (resp) => new ethers.providers.JsonRpcProvider(`https://rpc.tenderly.co/fork/${resp.data.simulation_fork.id}`)
+      )
+      .catch((err) => {
+        throw new Error(err)
+      })
```
willpote, line 120: Use async/await syntax

lib/handlers/tenderly-sim/tenderly-sim.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    log.info({ simulatedTxReceipt: swapReceipt }, 'Simulated Tx Via Tenderly')
+
+    return swapReceipt
+  }
+
+  private async getFork(params: { chainId: number; blockNumber: number }): Promise<JsonRpcProvider> {
+    log.info({ params: params }, 'Getting Tenderly Fork')
+    // Assume Fork of the block does not yet exist in our tenderly project
+    // TODO implement storage and management of tenderly forks
```
willpote, line 96: Should implement this. 

I think it would be fine to cache the forks for several blocks, so the cache should have that funcationality, and make the number the of blocks tunable in the constructor

lib/handlers/tenderly-sim/tenderly-sim.ts
```diff
@@ -0,0 +1,122 @@
...
+  TENDERLY_BASE_URL: string,
+  TENDERLY_USER: string,
+  TENDERLY_PROJECT: string
+) => `${TENDERLY_BASE_URL}/api/v1/account/${TENDERLY_USER}/project/${TENDERLY_PROJECT}/fork`
+
+export const APPROVE_TOKEN_FOR_TRANSFER =
+  '0x095ea7b300000000000000000000000068b3465833fb72a70ecdf485e0e4c7bd8665fc45ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff'
+
+export interface ISimulator {
+  simulateTx: (
```
willpote, line 23: simulateTransaction

lib/handlers/tenderly-sim/tenderly-sim.ts
```diff
@@ -0,0 +1,122 @@
...
+export const POST_TENDERLY_FORK_API_URL = (
+  TENDERLY_BASE_URL: string,
+  TENDERLY_USER: string,
+  TENDERLY_PROJECT: string
+) => `${TENDERLY_BASE_URL}/api/v1/account/${TENDERLY_USER}/project/${TENDERLY_PROJECT}/fork`
+
+export const APPROVE_TOKEN_FOR_TRANSFER =
+  '0x095ea7b300000000000000000000000068b3465833fb72a70ecdf485e0e4c7bd8665fc45ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff'
+
+export interface ISimulator {
```
willpote, line 22: Move to separate file

lib/handlers/quote/quote.ts
```diff
@@ -153,10 +156,14 @@ export class QuoteHandler extends APIGLambdaHandler<
         recipient: recipient,
         slippageTolerance: slippageTolerancePercent,
       }
+    } else {
+      // Can't simulate without swapParams set
+      simulate = false
     }
```
willpote, line 162: Should just update the Joi validation to make the request fail if simulate is set, but swap params aren't

willpote (commented):
Good progress! 

High level feedback- we should build this logic in the SOR, and if the simulation is requested we should make it replace all the other gas estimates we return. 

Theres no reason for us to return any of the other (less accurate) gas estimates, if we have run a simulation on Tenderly. 

Will also need the TODOs and tests implemented before merging

PR #64, opened 2022-07-21
Add differents alarms by severity

willpote:
Sev2s will page people with high urgency, then escalate if no response
Sev3s will just notify the oncall (via push notifications to phone/email)

Migrated all the existing alarms to be Sev3s, added new alarms for Sev2

### DIFF
--- a/bin/stacks/routing-api-stack.ts
+++ b/bin/stacks/routing-api-stack.ts
@@ -189,7 +189,40 @@ export class RoutingAPIStack extends cdk.Stack {
     })
     quoteToRatio.addMethod('GET', routeToRatioLambdaIntegration)
 
-    const apiAlarm5xx = new aws_cloudwatch.Alarm(this, 'RoutingAPI-5XXAlarm', {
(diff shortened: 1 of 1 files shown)


bin/stacks/routing-api-stack.ts
```diff
@@ -189,7 +189,40 @@ export class RoutingAPIStack extends cdk.Stack {
     })
     quoteToRatio.addMethod('GET', routeToRatioLambdaIntegration)
 
-    const apiAlarm5xx = new aws_cloudwatch.Alarm(this, 'RoutingAPI-5XXAlarm', {
+    const apiAlarm5xxSev2 = new aws_cloudwatch.Alarm(this, 'RoutingAPI-SEV2-5XXAlarm', {
```
snreynolds, line 192: Maybe we just leave comment that the comparison operator is default `GreaterThanOrEqualToThreshold`. Was just trying to figure out if the thresholds made sense for the diff severities but didnt know how the numbers were being compared

snreynolds (approved):

PR #66, opened 2022-07-21
Add cron for cleaning old IPFS pool caches

willpote:
Plan is to push it out with dryRun set to true, check logs, then remove the flag in a subsequent PR

One note, the Lambda may timeout (lambdas have max 15 min duration). I think this is fine, we can just let it timeout, and schedule the lambda to run a few times a week


snreynolds (approved):
lgtm!

PR #83, opened 2022-08-19
Added support for simulation

raghava-pamula:
Bumped SOR, added simulation support


lib/handlers/injector-sor.ts
```diff
@@ -237,6 +240,17 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
 
         const v2PoolProvider = new V2PoolProvider(chainId, multicall2Provider)
 
+        let simulator: ISimulator
+        simulator = new FallbackTenderlySimulator(
```
willpote, line 244: `const`?

lib/handlers/quote/schema/quote-schema.ts
```diff
@@ -47,6 +47,7 @@ export const QuoteQueryParamsJoi = Joi.object({
   forceCrossProtocol: Joi.boolean().optional(),
   forceMixedRoutes: Joi.boolean().optional(),
   protocols: Joi.stringArray().items(Joi.string().valid('v2', 'v3', 'mixed')).optional(),
+  fromAddress: Joi.string().optional(),
```
willpote, line 50: Use the same validation as the other addresses.

Also I think we should call this `simulateFromAddress`. Its not clear from just `fromAddress` that this would cause a simulation to be run.

bin/app.ts
```diff
@@ -110,6 +119,17 @@ export class RoutingAPIPipeline extends Stack {
       //secretCompleteArn: arn:aws:secretsmanager:us-east-2:644039819003:secret:routing-api-rpc-urls-json-backup-D2sWoe
     })
 
+    const tenderlyUser = sm.Secret.fromSecretAttributes(this, 'TenderlyUser', {
+      secretCompleteArn: '',
```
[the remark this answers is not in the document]
raghava-pamula: I think this may be an earlier commit, I added the secret and updated the code with it already

lib/handlers/quote/quote.ts
```diff
@@ -258,6 +262,7 @@ export class QuoteHandler extends APIGLambdaHandler<
       gasPriceWei,
       methodParameters,
       blockNumber,
+      simulationError,
     } = swapRoute
 
     const routeResponse: Array<(V3PoolInRoute | V2PoolInRoute)[]> = []
```
willpote, line 268: Lets emit some metrics on this so we can track usage. One for simulation requested, one for success and one for failure. See L86 for an example

willpote (approved):
Approving with 1 comment. Feel free to address then squash merge

PR #93, opened 2022-09-22
Update alarm thresholds, add new metrics+alarms for each individual chain, comment out goerli integ-tests

raghava-pamula:
* Lowered threshold for 5XX and Latency alarms so they get triggered more
* Increased threshold for poolcachetolambda and poolcachetoipfs alarms so they get triggered less
* Added metrics and Sev3 alarms for each chain for 200 Error % and 400 Error % 
* Improved tenderly simulation metrics and alarms
* Gorli integ tests failing due to lack of liquidity, temporarily commented it out

https://uniswaplabs.atlassian.net/jira/software/projects/TRD/boards/23?selectedIssue=TRD-105

### DIFF
--- a/bin/stacks/routing-api-stack.ts
+++ b/bin/stacks/routing-api-stack.ts
@@ -268,7 +268,7 @@ export class RoutingAPIStack extends cdk.Stack {
     const simulationAlarmSev2 = new aws_cloudwatch.Alarm(this, 'RoutingAPI-SEV2-Simulation', {
       alarmName: 'RoutingAPI-SEV2-Simulation',
       metric: new MathExpression({
-        expression: "100*(simulationFailed/simulationRequested)",
+        expression: '100*(simulationFailed/simulationRequested)',
         usingMetrics: {
           simulationRequested: new aws_cloudwatch.Metric({
             namespace: 'Uniswap',
@@ -276,7 +276,7 @@ export class RoutingAPIStack extends cdk.Stack {
             dimensionsMap: { Service: 'RoutingAPI' },
             unit: aws_cloudwatch.Unit.COUNT,
             period: Duration.minutes(5),
-            statistic: 'sum'
+            statistic: 'sum',
           }),
           simulationFailed: new aws_cloudwatch.Metric({
             namespace: 'Uniswap',
@@ -284,9 +284,9 @@ export class RoutingAPIStack extends cdk.Stack {
             dimensionsMap: { Service: 'RoutingAPI' },
             unit: aws_cloudwatch.Unit.COUNT,
             period: Duration.minutes(5),
-            statistic: 'sum'
+            statistic: 'sum',
           }),
-        }
+        },
       }),
       threshold: 40,
       evaluationPeriods: 2,
@@ -295,7 +295,7 @@ export class RoutingAPIStack extends cdk.Stack {
     const simulationAlarmSev3 = new aws_cloudwatch.Alarm(this, 'RoutingAPI-SEV3-Simulation', {
       alarmName: 'RoutingAPI-SEV3-Simulation',
       metric: new MathExpression({
-        expression: "100*(simulationFailed/simulationRequested)",
+        expression: '100*(simulationFailed/simulationRequested)',
         usingMetrics: {
           simulationRequested: new aws_cloudwatch.Metric({
             namespace: 'Uniswap',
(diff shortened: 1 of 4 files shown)


rileydcampbell (approved):

lib/handlers/quote/quote.ts
```diff
@@ -63,6 +63,7 @@ export class QuoteHandler extends APIGLambdaHandler<
         metric,
       },
     } = params
+    metric.putMetric(`GET_Quote_Requested_ChainId: ${chainId}`, 1, MetricLoggerUnit.Count)
```
rileydcampbell, line 66: Should we make these metrics all caps? Not sure if we are following a pattern here.
```suggestion
    metric.putMetric(`GET_QUOTE_REQUESTED_CHAIN_ID: ${chainId}`, 1, MetricLoggerUnit.Count)
```

bin/stacks/routing-api-stack.ts
```diff
@@ -259,22 +261,123 @@ export class RoutingAPIStack extends cdk.Stack {
         period: Duration.minutes(5),
         statistic: 'p90',
       }),
-      threshold: 7500,
+      threshold: 5500,
```
rileydcampbell, line 264: Would this make the alarm more sensitive?

bin/stacks/routing-api-stack.ts
```diff
@@ -206,7 +208,7 @@ export class RoutingAPIStack extends cdk.Stack {
         // For this metric 'avg' represents error rate.
         statistic: 'avg',
       }),
-      threshold: 0.25,
+      threshold: 0.05,
```
ConjunctiveNormalForm, line 211: wow this is pretty aggresive - wonder how you come up with this number?

bin/stacks/routing-api-stack.ts
```diff
@@ -259,22 +261,123 @@ export class RoutingAPIStack extends cdk.Stack {
...
     const simulationAlarmSev2 = new aws_cloudwatch.Alarm(this, 'RoutingAPI-SEV2-Simulation', {
       alarmName: 'RoutingAPI-SEV2-Simulation',
-      metric: api.metric('SimulationFailed'),
-      threshold: 3,
-      evaluationPeriods: 3,
+      metric: new MathExpression({
+        expression: '100*(simulationFailed/simulationRequested)',
+        usingMetrics: {
+          simulationRequested: new aws_cloudwatch.Metric({
+            namespace: 'Uniswap',
```
ConjunctiveNormalForm, line 274: does it make sense to use a more specific namespace?

raghava-pamula:
Screen Shot 2022-10-04 at 3 39 12 PM
The latency Sev3 alarm has only ever been triggered during extreme latency spikes (7500ms for Sev3, which IMO should be considered Sev2). I lowered it to 5500ms for Sev3 and 8500 for Sev2, so it should be triggered more

raghava-pamula:
Screen Shot 2022-10-04 at 3 41 48 PM
The 5XX Sev3 alarm only triggers during spikes. Sev3 alarm should alert us before spike, a spike like this with 5% 5XX rate should be Sev2 incident.

bin/stacks/routing-api-stack.ts
raghava-pamula: https://github.com/Uniswap/routing-api/pull/93#issuecomment-1267498466

bin/stacks/routing-api-stack.ts
raghava-pamula: that's right

lib/handlers/quote/quote.ts
raghava-pamula: Yeah I can change these

bin/stacks/routing-api-stack.ts
raghava-pamula: yes, but it is a bit more complicated change than you think. Here, it is not actually creating the metric, it is just retrieving the metric from the namespace. The namespace is actually set here: https://github.com/Uniswap/routing-api/blob/88eb9bef1f3e6f164791414ba2ae0dfc5ad7a295/lib/handlers/quote/injector.ts#L54
And the same namespace is used for all the custom metrics in routing-api. Good idea to split up the namespace, but it should be a separate PR imo

lib/handlers/quote/quote.ts
raghava-pamula: I made them all caps

ConjunctiveNormalForm (approved):

rileydcampbell (approved):

PR #98, opened 2022-10-07
Split PoolCacheToS3Lambda to multiple lambdas by chain and protocol

raghava-pamula:
https://uniswaplabs.atlassian.net/jira/software/projects/TRD/boards/23?selectedIssue=TRD-105
### Motivation:
https://uniswapteam.slack.com/archives/C02D07EM805/p1665108679405439 alarms like this are flooded eng ops. This is happening because V2 subgraph query keeps timing out and causes the poolCacheLambda to time out, which makes it seem that the whole cron failed, even though it worked for all other chains. After the lambda times out, it is retried. The cloudwatch rule is also configured to run this lambda once every 15 minutes. Because of this predictable failure on V2 and the retry policy, this lambda ends up stacking and running 4x concurrently, and spamming eng-ops with alerts.
Screen Shot 2022-10-06 at 2 07 32 PM
Screen Shot 2022-10-06 at 11 33 46 AM

### Solution:
Separate the pool cache lambda into multiple lambdas, one for each ChainId - Protocol pairing. Remove the pool cache lambda from the routing dashboard stack since it no longer exists. Separate the alerts. Reduce retries for V2 because the way its currently set up (10 minutes * 3 retries = 30 minutes) is double the lambda maximum duration of 15 minutes. Increase the timeout for V2 to 12 minutes to give it some extra time for the 1 and only retry it has. Change the Mainnet V2 alarm to be triggered if more than 85 percent of invocations error in an hour for 5 hours in a row (5 minute intervals) since it times out so often. Change all other alarms to be triggered if more than 50 percent of invocations error for 1 hour in a row.

Screen Shot 2022-10-07 at 1 17 38 AM

### Testing:
Chain and Protocol specific lambdas and alarms are created locally. V2 alarm still triggers, but all other ChainId-Protocol pairings work fine.

Screen Shot 2022-10-07 at 9 05 06 AM

Screen Shot 2022-10-07 at 2 11 24 AM

Also, removed deprecated testnets rinkeby and goerli.

### DIFF
--- a/bin/stacks/routing-caching-stack.ts
+++ b/bin/stacks/routing-caching-stack.ts
@@ -1,3 +1,4 @@
+import { Protocol } from '@uniswap/router-sdk'
 import * as cdk from 'aws-cdk-lib'
 import { Duration } from 'aws-cdk-lib'
 import * as aws_cloudwatch from 'aws-cdk-lib/aws-cloudwatch'
@@ -129,8 +130,8 @@ export class RoutingCachingStack extends cdk.NestedStack {
             }),
           },
         }),
-        threshold: 75,
-        evaluationPeriods: 1,
+        threshold: protocol===Protocol.V3 ? 50 : 85,
+        evaluationPeriods: protocol===Protocol.V3 ? 1 : 5,
       })
       const lambdaThrottlesErrorRate = new aws_cloudwatch.Alarm(this, `RoutingAPI-PoolCacheToS3LambdaThrottles-ChainId${chainId}-Protocol${protocol}`, {
         metric: lambda.metricThrottles({
--- a/bin/stacks/routing-caching-stack.ts
+++ b/bin/stacks/routing-caching-stack.ts
@@ -1,4 +1,3 @@
-import { Protocol } from '@uniswap/router-sdk'
 import * as cdk from 'aws-cdk-lib'
 import { Duration } from 'aws-cdk-lib'
 import * as aws_cloudwatch from 'aws-cdk-lib/aws-cloudwatch'
@@ -116,6 +115,23 @@ export class RoutingCachingStack extends cdk.NestedStack {
         targets: [new aws_events_targets.LambdaFunction(lambda)],
       })
       this.poolCacheBucket2.grantReadWrite(lambda)
+      const lambdaAlarmErrorRate = new aws_cloudwatch.Alarm(this, `RoutingAPI-PoolCacheToS3LambdaErrorRate-ChainId${chainId}-Protocol${protocol}`, {
+        metric: new MathExpression({
+          expression: '100*(errors/invocations)',
+          usingMetrics: {
+            invocations: lambda.metricInvocations({
+              period: Duration.minutes(60),
+              statistic: 'sum',
+            }),
+            errors: lambda.metricErrors({
+              period: Duration.minutes(60),
+              statistic: 'sum',
+            }),
+          },
+        }),
+        threshold: 50,
+        evaluationPeriods: 1,
+      })
       const lambdaThrottlesErrorRate = new aws_cloudwatch.Alarm(this, `RoutingAPI-PoolCacheToS3LambdaThrottles-ChainId${chainId}-Protocol${protocol}`, {
         metric: lambda.metricThrottles({
           period: Duration.minutes(5),
@@ -125,27 +141,8 @@ export class RoutingCachingStack extends cdk.NestedStack {
         evaluationPeriods: 1,
       })
       if (chatBotTopic) {
+        lambdaAlarmErrorRate.addAlarmAction(new aws_cloudwatch_actions.SnsAction(chatBotTopic))
         lambdaThrottlesErrorRate.addAlarmAction(new aws_cloudwatch_actions.SnsAction(chatBotTopic))
-        if(protocol !== Protocol.V2) {
-          const lambdaAlarmErrorRate = new aws_cloudwatch.Alarm(this, `RoutingAPI-PoolCacheToS3LambdaErrorRate-ChainId${chainId}-Protocol${protocol}`, {
-            metric: new MathExpression({
-              expression: '100*(errors/invocations)',
-              usingMetrics: {
-                invocations: lambda.metricInvocations({
-                  period: Duration.minutes(60),
-                  statistic: 'sum',
-                }),
-                errors: lambda.metricErrors({
-                  period: Duration.minutes(60),
-                  statistic: 'sum',
-                }),
-              },
-            }),
-            threshold: 50,
-            evaluationPeriods: 1,
-          })
-          lambdaAlarmErrorRate.addAlarmAction(new aws_cloudwatch_actions.SnsAction(chatBotTopic))
-        }
       }
     }
 
--- a/bin/stacks/routing-caching-stack.ts
+++ b/bin/stacks/routing-caching-stack.ts
@@ -131,7 +131,7 @@ export class RoutingCachingStack extends cdk.NestedStack {
           },
         }),
         threshold: protocol===Protocol.V3 ? 50 : 85,
-        evaluationPeriods: protocol===Protocol.V3 ? 1 : 5,
+        evaluationPeriods: protocol===Protocol.V3 ? 6 : 30,
       })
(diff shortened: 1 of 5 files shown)


bin/stacks/routing-caching-stack.ts
```diff
@@ -37,6 +39,11 @@ export class RoutingCachingStack extends cdk.NestedStack {
 
     const { chatbotSNSArn } = props
 
+    let chatBotTopic: cdk.aws_sns.ITopic | undefined
+    if (chatbotSNSArn) {
+      chatBotTopic = aws_sns.Topic.fromTopicArn(this, 'ChatbotTopic', chatbotSNSArn)
+    }
+
```
ConjunctiveNormalForm, line 46: nit: `const chatBotTopic = chatbotSNSArn ? aws_sns.Topic.fromTopicArn(this, 'ChatbotTopic', chatbotSNSArn) : undefined`

ConjunctiveNormalForm (commented):
>Reduce retries for V2 because the way its currently set up (10 minutes * 3 retries = 30 minutes) is double the lambda maximum duration of 15 minutes. 

Are we sure lambda retries don't count as separate invocations? (i.e. each retry has it's own 15-minute limit)?

bin/stacks/routing-caching-stack.ts
raghava-pamula: > > Reduce retries for V2 because the way its currently set up (10 minutes * 3 retries = 30 minutes) is double the lambda maximum duration of 15 minutes.
> 
> Are we sure lambda retries don't count as separate invocations? (i.e. each retry has it's own 15-minute limit)?

sorry when I said this i didn't mean lambda retry, I meant subgraph retry: https://github.com/Uniswap/smart-order-router/blob/d0fdfb618656476bacfef802a21df51390614a40/src/providers/v2/subgraph-provider.ts#L69

raghava-pamula:
> > Reduce retries for V2 because the way its currently set up (10 minutes * 3 retries = 30 minutes) is double the lambda maximum duration of 15 minutes.
> 
> Are we sure lambda retries don't count as separate invocations? (i.e. each retry has it's own 15-minute limit)?

sorry when I said this i didn't mean lambda retry, I meant subgraph retry: https://github.com/Uniswap/smart-order-router/blob/d0fdfb618656476bacfef802a21df51390614a40/src/providers/v2/subgraph-provider.ts#L69

ConjunctiveNormalForm:
if the v2 provider times out at 12 min, does it mean that it only leaves 3 mins for it to retry before the entire lambda times out?

bin/stacks/routing-dashboard-stack.ts
```diff
@@ -401,30 +400,6 @@ export class RoutingDashboardStack extends cdk.NestedStack {
...
-                      ['AWS/Lambda', 'Errors', 'FunctionName', ipfsPoolCacheLambdaName],
-                      ['.', 'Invocations', '.', '.'],
-                    ]
-                  : []),
-              ],
-              region: region,
-              title: 'Pool Cache Lambda Error/Invocations | 5min',
-              stat: 'Sum',
-            },
-          },
```
willpote, line 427: Why remove the dashboard entirely? We could just have the same dashboard, but with lines for each chain

lib/cron/cache-pools.ts
```diff
@@ -1,11 +1,17 @@
...
 import { S3_POOL_CACHE_KEY } from '../util/pool-cache-key'
-import { chainProtocols } from './cache-config'
 
 const handler: ScheduledHandler = async (event: EventBridgeEvent<string, void>) => {
+  const chainId: ChainId = parseInt(process.env.chainId!)
+  const protocol = process.env.protocol! as Protocol
+  const timeout = parseInt(process.env.timeout!)
+  // Don't retry for V2 as it will timeout and throw 500
+  const provider =
+    protocol === Protocol.V3 ? new V3SubgraphProvider(chainId, 3, timeout) : new V2SubgraphProvider(chainId, 1, timeout)
```
willpote, line 14: Why do we reinitialize the subgraph provider here instead of just using the providers created in cache-config? It would be simpler to just pass in chainId and protocol and do a lookup from cache-config?

bin/stacks/routing-caching-stack.ts
```diff
@@ -72,40 +79,85 @@ export class RoutingCachingStack extends cdk.NestedStack {
...
+                statistic: 'sum',
+              }),
+              errors: lambda.metricErrors({
+                period: Duration.minutes(60),
+                statistic: 'sum',
+              }),
+            },
+          }),
+          threshold: protocol === Protocol.V3 ? 50 : 85,
+          evaluationPeriods: protocol === Protocol.V3 ? 12 : 60,
```
willpote, line 141: Are you sure this is set correctly? My understand is that evaluation periods is set based on the metric field, so since we specified period as "60 minutes" above, then this would be 12 * 60 minutes so 12 hours no?

raghava-pamula:
> if the v2 provider times out at 12 min, does it mean that it only leaves 3 mins for it to retry before the entire lambda times out?

I changed it to 14 minutes and 0 retry policy, so it will just throw here instead of timing out: https://github.com/Uniswap/routing-api/blob/c32e6100821478aa49122988bdf7d309fc3dadf5/lib/cron/cache-pools.ts#L27

bin/stacks/routing-caching-stack.ts
raghava-pamula: the evaluation period is 5 minutes, so its 12 * 5 minutes = 1 hour

bin/stacks/routing-dashboard-stack.ts
raghava-pamula: I'll add back

lib/cron/cache-pools.ts
```diff
@@ -15,40 +24,36 @@ const handler: ScheduledHandler = async (event: EventBridgeEvent<string, void>)
...
-      } catch (err) {
-        log.error({ err }, `Failed to get pools for ${protocol} on ${chainId}`)
-        throw new Error(`Failed to get pools for ${protocol} on ${chainId}`)
-      }
+  let pools
+  try {
+    pools = await provider.getPools()
+  } catch (err) {
+    log.error({ err }, `Failed to get pools for ${protocol} on ${chainId}`)
+    throw new Error(`Failed to get pools for ${protocol} on ${chainId}`)
```
rileydcampbell, line 34: Because we set a lower timeout than 15 mins if this fails should it no longer post in #eng-ops? Should we add a CloudWatch metric here to log?

raghava-pamula: This will post to eng ops since it throws

willpote (approved):

bin/stacks/routing-dashboard-stack.ts
```diff
@@ -76,11 +85,40 @@ export class RoutingDashboardStack extends cdk.NestedStack {
       },
     })
 
+    const poolCacheLambdaWidgetArray: LambdaWidget[] = []
+    poolCacheLambdaNameArray.forEach((poolCacheLambdaName) => {
```
ConjunctiveNormalForm, line 89: afaict this creates multiple widgets instead of creating one with metric (time series data) for each chain, right? feel like the latter is cleaner (and is also what Pote originally suggested)

raghava-pamula: Oh I'm not sure how to do that, can you give a suggestion? I'm also looking

raghava-pamula: think i figured it out

ConjunctiveNormalForm: cool, I was thinking adding 
```
            ['AWS/Lambda', 'Errors', 'FunctionName', poolCacheLambdaName],
            ['.', 'Invocations', '.', '.'],
```
for each `poolCacheLambdaName` in `poolCacheLambdaNameArray` to the `metrics` array field

raghava-pamula: Ok updated it

ConjunctiveNormalForm (approved):

lib/cron/cache-pools.ts
rileydcampbell: So in the future would we want to just log these error as a cloudwatch metric instead of throwing an error? Then we could have an alarm that says if we get X error metrics over Y amount of time then page?

rileydcampbell: oh jk this is a categorization issue. ignore

PR #100, opened 2022-10-10
change cache pools ipfs lambda to be run in parallel for each chain/p…

raghava-pamula:
change cache pools ipfs lambda to be run in parallel for each chain/protocol pairing. Before, the handler run sequentially for each pairing, this should be faster and less likely to time out.

### DIFF
--- a/bin/stacks/routing-caching-stack.ts
+++ b/bin/stacks/routing-caching-stack.ts
@@ -235,7 +235,7 @@ export class RoutingCachingStack extends cdk.NestedStack {
             period: Duration.minutes(60),
             statistic: 'sum',
           }),
-          threshold: 9,
+          threshold: 13,
           evaluationPeriods: 1,
         })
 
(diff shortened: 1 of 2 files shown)


bin/stacks/routing-caching-stack.ts
```diff
@@ -235,7 +235,7 @@ export class RoutingCachingStack extends cdk.NestedStack {
             period: Duration.minutes(60),
             statistic: 'sum',
           }),
-          threshold: 9,
+          threshold: 13,
```
rileydcampbell, line 238: Do you know roughly how many times this is failing in the 60 min period currently? Would making this 13 be a high enough threshold?

rileydcampbell (approved):
lgtm, just one question.

ConjunctiveNormalForm (approved):

bin/stacks/routing-caching-stack.ts
raghava-pamula: right now it is failing at most 11 times in an hour. 13 should not be hit unless something is really wrong

PR #101, opened 2022-10-17
Fix pool cache lambda

raghava-pamula:
This fixes the root cause of ongoing issue with pool cache lambda where V2 Pool Subgraph Query is failing. The root cause was that pageSize was 1000, which causes subgraph query to timeout. This PR changes the pageSize to 250. When I deployed it locally, the lambda does not fail anymore. The query fails sometimes, but retries almost always succeed. Also, this configures SOR logging to show up in cloudwatch. Before, SOR logs were not showing up.
Screen Shot 2022-10-17 at 12 46 40 PM


lib/cron/cache-pools.ts
```diff
@@ -19,6 +19,7 @@ const handler: ScheduledHandler = async (event: EventBridgeEvent<string, void>)
     level: 'info',
     requestId: event.id,
   })
+  setGlobalLogger(log)
```
rileydcampbell, line 22: What will this do?

rileydcampbell (approved):

lib/cron/cache-pools.ts
raghava-pamula: it configures the SOR logs to show up in cloudwatch. Before they weren't showing up, pretty sure someone just forgot to add this

ConjunctiveNormalForm (approved):

PR #102, opened 2022-10-17
Update simulation alarm threshold criterion

raghava-pamula:
Before, even one failed simulation triggered the alarm because the feature is not used that much yet. Now, 3/5 periods of 5 minutes each must have error rate of 20% before the alarm is triggered.


bin/stacks/routing-api-stack.ts
```diff
@@ -299,7 +299,8 @@ export class RoutingAPIStack extends cdk.Stack {
         },
       }),
       threshold: 40,
-      evaluationPeriods: 2,
+      evaluationPeriods: 3,
+      datapointsToAlarm: 2
```
ConjunctiveNormalForm: this makes the alarm even more sensitive than before, no?
previously it needs two consecutive periods evaluate to over threshold, now it only needs any two out of the latest three.

raghava-pamula: yeah you're right

raghava-pamula: changed to evaulationPeriods = 5 and datapointsToAlarm = 4

bin/stacks/routing-api-stack.ts
```diff
@@ -326,7 +327,8 @@ export class RoutingAPIStack extends cdk.Stack {
         },
       }),
       threshold: 20,
-      evaluationPeriods: 2,
+      evaluationPeriods: 5,
+      datapointsToAlarm: 4,
```
ConjunctiveNormalForm, line 331: SEV3 should probably be lower? 3 out of 5?

raghava-pamula: changed it to 3 out of 5

ConjunctiveNormalForm (approved):

rileydcampbell (approved):

raghava-pamula:
https://github.com/Uniswap/routing-api/commit/5b5322d11a214d67e8c50f0996abf7229e7feace
zach addressed it here

PR #105, opened 2022-10-27
Adjust the expression for caching alarm.

rileydcampbell:
## Summary
Changing the expression for the pool cache lambda alarms to check if it has succeeded at least once in the interval.

### DIFF
--- a/bin/stacks/routing-caching-stack.ts
+++ b/bin/stacks/routing-caching-stack.ts
@@ -123,7 +123,7 @@ export class RoutingCachingStack extends cdk.NestedStack {
         `RoutingAPI-SEV4-PoolCacheToS3LambdaErrorRate-ChainId${chainId}-Protocol${protocol}`,
         {
           metric: new MathExpression({
-            expression: '100*(errors/invocations)',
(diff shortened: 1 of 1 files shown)


raghava-pamula (approved):

ConjunctiveNormalForm:
what happens when there is no invocation?

rileydcampbell:
> what happens when there is no invocation?
@raghava-pamula correct me if I'm wrong but I think this runs on a schedule so invocations will never be zero. If they were zero then this would be triggering in the current state? cc: @ConjunctiveNormalForm

ConjunctiveNormalForm:
> > what happens when there is no invocation?
> 
> @raghava-pamula correct me if I'm wrong but I think this runs on a schedule so invocations will never be zero. If they were zero then this would be triggering in the current state? cc: @ConjunctiveNormalForm

that makes sense; I thought about this because I remember Dan mentioning that sometimes cron jobs will just straight up not trigger, but I guess that's a rare enough situation to not worry about, especially when the evaluation period is pretty high

ConjunctiveNormalForm (approved):

PR #112, opened 2022-12-12
bump SOR, use ethEstimateGas simulator when 'skipTenderlySimulation' flag is passed

raghava-pamula:
* This PR updates the gas fee flow to use ethEstimateGasSimulator when the skipTenderlySimulation flag is passed in.
* When the skipTenderlySimulation flag is not present, the fallback tenderlySimulator is used.
* The plan is for Web to integrate with this new gas fee flow by passing in simulateFromAddress and skipTenderlySimulation=true. This way, Web will get gas limit estimates using ethEstimateGas when possible, but never through Tenderly.
* Added integration tests for the skipTenderlySimulation flag.

### DIFF
--- a/lib/handlers/quote/quote.ts
+++ b/lib/handlers/quote/quote.ts
@@ -456,7 +456,7 @@ export class QuoteHandler extends APIGLambdaHandler<
       gasUseEstimate: estimatedGasUsed.toString(),
       gasUseEstimateUSD: estimatedGasUsedUSD.toExact(),
       simulationStatus: simulationStatus ? simulationStatusToString(simulationStatus, log) : undefined,
-      simulationError: (simulationStatus === undefined || simulationStatus == SimulationStatus.Succeeded) ? false : true,
+      simulationError: simulationStatus === undefined || simulationStatus == SimulationStatus.Succeeded ? false : true,
       gasPriceWei: gasPriceWei.toString(),
       route: routeResponse,
       routeString: routeAmountsToString(route),
--- a/lib/handlers/quote/quote.ts
+++ b/lib/handlers/quote/quote.ts
@@ -348,7 +348,7 @@ export class QuoteHandler extends APIGLambdaHandler<
     } else if (simulationStatus == SimulationStatus.NotApproved) {
       metric.putMetric('SimulationNotApproved', 1, MetricLoggerUnit.Count)
     } else if (simulationStatus == SimulationStatus.NotSupported) {
-        metric.putMetric('SimulationNotSupported', 1, MetricLoggerUnit.Count)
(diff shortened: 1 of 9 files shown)


lib/handlers/quote/quote.ts
```diff
@@ -453,8 +455,8 @@ export class QuoteHandler extends APIGLambdaHandler<
       gasUseEstimateQuoteDecimals: estimatedGasUsedQuoteToken.toExact(),
       gasUseEstimate: estimatedGasUsed.toString(),
       gasUseEstimateUSD: estimatedGasUsedUSD.toExact(),
-      simulationStatus: simulationStatusToString(simulationStatus, log),
-      simulationError: (simulationStatus == SimulationStatus.Succeeded || simulationStatus == SimulationStatus.Unattempted) ? false : true,
+      simulationStatus: simulationStatus ? simulationStatusToString(simulationStatus, log) : undefined,
+      simulationError: simulationStatus === undefined || simulationStatus == SimulationStatus.Succeeded ? false : true,
```
rileydcampbell, line 459: Just trying to get some more context here because I'm not super familiar with this code. Why would `simulationError` be `false` if `simulationStatus` is `undefined`? Is that because we are skipping the simulation so we are okay with the undefined status?

lib/handlers/quote/util/simulation.ts
```diff
@@ -1,16 +1,20 @@
...
     case SimulationStatus.Failed:
       return 'FAILED'
-    case SimulationStatus.Unattempted:
-      return 'UNATTEMPTED'
+    case SimulationStatus.NotApproved:
+      return 'NOTAPPROVED'
+    case SimulationStatus.InsufficientBalance:
+      return 'INSUFFICIENTBALANCE'
+    case SimulationStatus.NotSupported:
+      return 'NOTSUPPORTED'
```
rileydcampbell, line 15: Nit: would it make sense for these enums to be snake case? Feel like that's a pattern I've seen us use elsewhere. 
```suggestion
    case SimulationStatus.NotApproved:
      return 'NOT_APPROVED'
    case SimulationStatus.InsufficientBalance:
      return 'INSUFFICIENT_BALANCE'
    case SimulationStatus.NotSupported:
      return 'NOT_SUPPORTED'
```

rileydcampbell (commented):
Don't have a ton of context on this part of the code but looks pretty good to me. Will take another more thorough look tomorrow.

lib/handlers/quote/quote.ts
raghava-pamula: yeah that's right, simulationStatus will be undefined if the client didn't request a simulation, so it is fine.

PR #114, opened 2023-02-08
Create a ChainId dimension for routing-api metrics

mikeki:
This is a metrics change where instead of encoding the ChainId in the metric name we instead create a dimension for the metrics.


willpote (approved):

mikeki:
This change is making the metrics behaves in ways that I didn't expect. I'm closing the branch for now, but I intend to come back to this as soon as I have to time to understand how the metrics and dimensions work.

PR #116, opened 2023-02-16
Add metrics to track amounts from a hardcoded list of pairs

mikeki:
Added 2 metrics in the routing api to keep track of the amounts and buckets that are currently being generated.

### DIFF
--- a/lib/handlers/quote/quote.ts
+++ b/lib/handlers/quote/quote.ts
@@ -12,6 +12,7 @@ import {
   SwapOptions,
   SwapType,
   SimulationStatus,
+  IMetric,
 } from '@uniswap/smart-order-router'
 import { Pool } from '@uniswap/v3-sdk'
 import JSBI from 'jsbi'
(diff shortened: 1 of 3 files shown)


lib/handlers/quote/quote.ts
```diff
@@ -459,17 +462,60 @@ export class QuoteHandler extends APIGLambdaHandler<
...
+    const PAIRS_TO_TRACK = [
+      'WETH/USDC',
+      'USDC/WETH',
+      'USDT/WETH',
+      'WETH/USDT',
+      'USDC/USDT',
+      'USDT/USDC',
+      'WMATIC/USDC',
+      'USDC/WMATIC',
+    ]
```
willpote, line 497: Its best to always use addresses since tokens can use the same symbol

willpote: The addresses can be different on different chains though, so need to handle that

mikeki: I think we are okay with that collision for now, these metrics are mostly to gather some initial data but I wouldn't keep them around forever.

Would it be okay if we kept this way for now, with the understanding that tokens could use the same symbol? I think the likelihood of that happening is low given the pairs being used.

willpote (approved):

PR #120, opened 2023-02-24
Addressing review feedback

mikeki:
In this branch I'm addressing the feedback recevied in the previous branch.

2 main changes:
1. Pais to track is now used to determine which metrics to log as well as which widgets to create for our dashboard. Eventually we might even use it to define our buckets and cache keys.
2. TradeType is now coming from sdk-core instead of redefining a new enum.


zhongeric (approved):
Lgtm!

PR #137, opened 2023-03-18
Create DynamoRouteCachingProvider and CachedRoutesConfiguration

mikeki:
### DynamoDB Route Caching Provider

This branch does the following:

1. Creates a nested stack which is used to define the DynamoDB database used for caching
2. Wires up the table name, and grants read write permission, to the lambda
3. Implements a `CachedRouteStrategy` and `CachedRouteParameters` class to define the buckets to be used for caching
4. Implements a `PairTradeTypeChainId` class to be used as a generator for the key to access the strategies
5. Implements the `DynamoRouteCachingProvider` which inherits from `IRouteCachingProvider`, in here we define how to fetch `CacheMode` and `BlocksToLive` coming from the strategy. We also decide to insert the `CachedRoutes` to DynamoDB based on the strategy and all the logic for inserting into DynamoDB is found here
6. There's also logic for fetching any existing `CachedRoute` from DynamoDB and returning the most recent entry. The calling function of `IRouteCachingProvider` is in charge of validating the blocksToLive setting is respected.
7. Given how complex each `CachedRoute` object is, the easiest solution was to store the object in binary, in order to do that I required to have full control of the class that gets serialized to binary so a Marshaller was created to extract fields from the object and to re-create the objects when reading from cache/binary.


lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
```diff
@@ -0,0 +1,302 @@
...
+    tradeType: TradeType,
+    protocols: Protocol[]
+  ): Promise<CachedRoutes | undefined> {
+    const [tokenIn, tokenOut] = this.determineTokenInOut(amount, quoteToken, tradeType)
+    const cachedRoutesStrategy = this.getCachedRoutesStrategy(tokenIn, tokenOut, tradeType, chainId)
+    const cachingParameters = cachedRoutesStrategy?.getCachingParameters(amount)
+
+    if (cachingParameters) {
+      const partitionKey = `${tokenIn.address}/${tokenOut.address}/${tradeType}/${chainId}`
+      const partialSortKey = `${protocols}/${cachingParameters.bucket}/`
```
zhongeric, line 77: Should we serialize the `protocols` array and add a deliminator? Also do you need the trailing `/`?

lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
```diff
@@ -0,0 +1,302 @@
...
+        ExpressionAttributeValues: {
+          ':pk': partitionKey,
+          ':sk': partialSortKey,
+        },
+        ScanIndexForward: false, // Reverse order to retrieve most recent item first
+        Limit: 1, // Only retrieve the most recent item
+      }
+
+      try {
+        log.info({ queryParams }, `[DynamoRouteCachingProvider] Attempting to get route from cache.`)
```
zhongeric, line 96: Is this `log` imported from smart order router? Shouldn't we be using the one in routing-api?

lib/handlers/router-entities/route-caching/marshalling/cached-route-marshaller.ts
```diff
@@ -0,0 +1,24 @@
...
+    }
+  }
+
+  public static unmarshal(marshalledCachedRoute: MarshalledCachedRoute): CachedRoute<V3Route | V2Route | MixedRoute> {
+    return new CachedRoute<V3Route | V2Route | MixedRoute>({
+      route: RouteMarshaller.unmarshal(marshalledCachedRoute.route),
+      percent: marshalledCachedRoute.percent,
+    })
+  }
+}
```
zhongeric, line 24: Move into `cached-routes-marshaller` ?

lib/handlers/router-entities/route-caching/marshalling/pair-marshaller.ts
```diff
@@ -0,0 +1,26 @@
+import { Pair } from '@uniswap/v2-sdk'
+import { CurrencyAmountMarshaller, MarshalledCurrencyAmount } from './currency-amount-marshaller'
+import { Protocol } from '@uniswap/router-sdk'
+
+export interface MarshalledPair {
+  protocol: Protocol
+  currencyAmountA: MarshalledCurrencyAmount
+  tokenAmountB: MarshalledCurrencyAmount
```
zhongeric, line 8: Why the different naming here between `currencyAmount` and `tokenAmount`

lib/handlers/router-entities/route-caching/cached-routes-configuration.ts
```diff
@@ -0,0 +1,50 @@
...
+      tradeType: TradeType.EXACT_INPUT,
+      chainId: ChainId.MAINNET,
+    }).toString(),
+    new CachedRoutesStrategy([
+      new CachedRoutesParameters({ bucket: 1, blocksToLive: 2, cacheMode: CacheMode.Tapcompare }),
+      new CachedRoutesParameters({ bucket: 5, blocksToLive: 2, cacheMode: CacheMode.Tapcompare }),
+      new CachedRoutesParameters({ bucket: 10, blocksToLive: 1, cacheMode: CacheMode.Tapcompare }),
+      new CachedRoutesParameters({ bucket: 50, blocksToLive: 1, cacheMode: CacheMode.Tapcompare }),
+      new CachedRoutesParameters({ bucket: 100, blocksToLive: 1, cacheMode: CacheMode.Tapcompare }),
+      new CachedRoutesParameters({ bucket: 500, blocksToLive: 1, cacheMode: CacheMode.Tapcompare }),
```
zhongeric, line 29: Coming back to this, these buckets seem far too large right? I.e. a 1.01 WETH -> USDC would hit the 5 WETH -> USDC one?

zhongeric (commented):
Nice! Left some notes

bin/stacks/routing-lambda-stack.ts
```diff
@@ -23,6 +24,7 @@ export interface RoutingLambdaStackProps extends cdk.NestedStackProps {
   tenderlyProject: string
   tenderlyAccessKey: string
   chatbotSNSArn?: string
+  cachedRoutesDynamoDb: aws_dynamodb.Table
```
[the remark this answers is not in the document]
mikeki: sure, I can make this optional. The main setup will deploy a DynamoDB table, so I think external people working on this api would need to modify that to left it out, but I think it's still a good idea to make this optional

lib/handlers/injector-sor.ts
```diff
@@ -288,6 +291,8 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
           })(),
         ])
 
+        const routeCachingProvider = new DynamoRouteCachingProvider({ cachedRoutesTableName: CACHED_ROUTES_DB! })
+
```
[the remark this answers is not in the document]
mikeki: good tip, thanks

lib/handlers/router-entities/route-caching/cached-routes-configuration.ts
```diff
@@ -0,0 +1,50 @@
...
+export const CACHED_ROUTES_CONFIGURATION: Map<string, CachedRoutesStrategy> = new Map([
+  [
+    new PairTradeTypeChainId({
+      tokenIn: '0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2', // WETH
+      tokenOut: '0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48', // USDC
+      tradeType: TradeType.EXACT_INPUT,
+      chainId: ChainId.MAINNET,
+    }).toString(),
+    new CachedRoutesStrategy([
+      new CachedRoutesParameters({ bucket: 1, blocksToLive: 2, cacheMode: CacheMode.Tapcompare }),
```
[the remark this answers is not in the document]
mikeki: yeah, I would need to change it there and then pull new version, I think at this point we rather keep it as is

lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
mikeki: I don't think we need to serialize it, it's an array and we will get something like: V2,V3,MIXED or V3 or V2,V3 and thats what I wanted to get, so protocols is fine, it's mostly a way to have different cache entries depending on the requested protocols.

The trailing `/` is to be able to match the last bit only which is the block number, we could keep it or remove it, but I opted for keeping it to make it obvious that we are just leaving the blockNumber out

lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
mikeki: I think the `log` comes from smart-order-router and it is injected from the injected-sor, so it is the same log

lib/handlers/router-entities/route-caching/marshalling/cached-route-marshaller.ts
mikeki: This is different than cached-routes-marshaller, they are 2 different objects.

lib/handlers/router-entities/route-caching/marshalling/pair-marshaller.ts
mikeki: that's the name that the `Pair` has in v2-sdk, so I just kept that consistent

lib/handlers/router-entities/route-caching/cached-routes-configuration.ts
mikeki: probably yeah, although I don't expect it to be too different. I pulled these number from the boundaries in the TrackedPairsDashboard from routing-api, but yeah we could create different buckets

lib/handlers/router-entities/route-caching/marshalling/route-marshaller.ts
```diff
@@ -0,0 +1,91 @@
...
+        pairs: route.pairs.map((pair) => PairMarshaller.marshal(pair)),
+      }
+    } else if (route.protocol === Protocol.V3) {
+      return {
+        protocol: Protocol.V3,
+        input: TokenMarshaller.marshal(route.input),
+        output: TokenMarshaller.marshal(route.output),
+        pools: route.pools.map((pool) => PoolMarshaller.marshal(pool)),
+      }
+    } else {
```
willpote, line 47: Nit: defensively code against adding a new protocol later by doing 

`else if (route.protocol === Protocol.MIXED)`

then 

`else { throw Error() }`

willpote (approved):
Mostly nits, looks good!

What is the rollout plan - wondering how we can reduce risk as we deploy this?

lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
```diff
@@ -0,0 +1,302 @@
...
+
+        return true
+      } catch (error) {
+        log.error({ error, putParams }, `[DynamoRouteCachingProvider] Cached route failed to insert`)
+
+        return false
+      }
+    } else {
+      // No CachingParameters found, return false to indicate the route was not cached.
+
```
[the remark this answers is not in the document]
mikeki: I don't think we need to log something here, this will be the "main" case since the CachingParameters are in the configuration file, I think that these logs will be unnecessary

lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
```diff
@@ -0,0 +1,303 @@
...
+}
+
+export class DynamoRouteCachingProvider extends IRouteCachingProvider {
+  private readonly ddbClient: DynamoDB.DocumentClient
+  private readonly tableName: string
+  private readonly ttlMinutes: number
+
+  constructor({ cachedRoutesTableName, ttlMinutes = 5 }: ConstructorParams) {
+    super()
+    this.ddbClient = new DynamoDB.DocumentClient()
```
willpote, line 29: I think the default retry strategy for AWS SDK calls is 3 retries with exponential backoff. Not sure what the default timeout is

Wondering if this is the best config for a cache? May want to fast fail with 0 retries? And timeout at 100ms or something?

mikeki: Updated :)

PR #140, opened 2023-04-04
Add objects to generate the partition and sort keys, and compare more routes

mikeki:
- Add objects for the partial key too
- Add a few more routes to compare


zhongeric (approved):
Looks great, thanks!

PR #142, opened 2023-04-05
Improve CachedRoutes Dashboard

mikeki:
- Update Dashboard to show the table of quotes with differences, and enable cache for low amounts
- Refactor cached routes configuration related code
- Refactor metrics dashboard of cached routes
- add test for cached routes widgets factory
- fix tests
- update comment
- it's not necessary to track gorli


lib/handlers/router-entities/route-caching/model/cached-routes-strategy.ts
```diff
@@ -1,32 +1,70 @@
...
+
+  public readablePairTradeTypeChainId(): string {
+    return `${this.pair.toUpperCase()}/${this.tradeType}/${this.chainId}`
+  }
+
+  public bucketPairs(): [number, number][] {
+    const firstBucket: [number, number][] = [[0, this.buckets[0]]]
+    const middleBuckets: [number, number][] = this.buckets
+      .slice(0, -1)
+      .map((bucket, i): [number, number] => [bucket, this.buckets[i + 1]!])
```
zhongeric, line 57: If `buckets` is of length 1 , `this.buckets[i + 1]!` will be out of range right? Not sure if the result of slice(0, -1) would have one element or no elements in this case though.

lib/handlers/router-entities/route-caching/model/cached-routes-strategy.ts
```diff
@@ -1,32 +1,70 @@
...
+    const lastBucket: [number, number][] = [[this.buckets.slice(-1)[0], -1]]
+
+    return firstBucket.concat(middleBuckets).concat(lastBucket)
   }
 
   /**
    * Given an amount, we will search the bucket that has a cached route for that amount based on the CachedRoutesParameters array
```
zhongeric, line 64: nit:  comment should say bucketsMap mapping instead of CachedRoutesParameters

lib/dashboards/cached-routes-widgets-factory.ts
```diff
@@ -68,31 +72,107 @@ export class CachedRoutesWidgetsFactory implements WidgetsFactory {
     ]
   }
 
-  private generateQuoteDiffWidgetsFromPair(pair: string): Widget[] {
-    return [
+  private generateQuoteDiffWidgetsFromPair(cacheStrategy: CachedRoutesStrategy): Widget[] {
+    const pairTradeTypeChainId = cacheStrategy.readablePairTradeTypeChainId()
+    const getQuoteMetricName = `GET_QUOTE_AMOUNT_${cacheStrategy.pair}_${cacheStrategy.tradeType.toUpperCase()}_CHAIN_${
+      cacheStrategy.chainId
+    }`
```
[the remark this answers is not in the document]
mikeki: I think those metrics are defined in SOR, these metrics are already defined in routing-api following the all caps pattern in routing-api quote lambda... I agree that we need to standardize it though, but it seems out of scope for this branch

lib/handlers/router-entities/route-caching/model/cached-routes-strategy.ts
mikeki: great catch, I fixed this, and also handled the case of 0 buckets.

zhongeric (approved):

PR #152, opened 2023-04-11
Update heuristics for selecting cached routes

mikeki:
- Add new heuristics for deciding when to add a route to cache
- Update more configuration


mikeki:
Self-approving and landing - this is a minor code change and mostly configuration change to caching

PR #155, opened 2023-04-13
Caching configurations for tracking new pairs

mikeki:
Adding new configurations of pairs to track and testing more generalized configurations


mikeki:
Self-approving. This is just a configuration for tracking purposes

PR #156, opened 2023-04-13
Increase blocks to live on arbitrum, since there are 3 blocks created per second

mikeki:


mikeki:
slef-approval. Dashboard and tracking changes

PR #161, opened 2023-04-18
Enabling USDT/WETH pairs for caching, and disable low amounts buckets of wildcards to test configuration

mikeki:


mikeki:
Selg-approve configuration change

PR #163, opened 2023-04-18
Iteration on wildcard buckets

mikeki:
- Revert "Darkmoding small quotes on wildcards (#162)"
- Another iteration on wildcard buckets


mikeki:
self-approve

PR #165, opened 2023-04-19
Allow the wildcard pairs amounts to be measured as well

mikeki:


mikeki:
self-approve dashboard change

PR #166, opened 2023-04-19
Updated cached routes logs dashboard

mikeki:
- Add more fields to the cached routes dashboard
- fix error


mikeki:
self-approve for dashboard change

PR #167, opened 2023-04-19
Add more pairs to track

mikeki:


mikeki:
self-approve simple dashboard change

PR #170, opened 2023-04-20
Update dashboard to have the min as 0 for better comparison

mikeki:


mikeki:
self-approve small dashboard config

PR #176, opened 2023-04-27
Service log level changed to Warn

mikeki:
We shouldn't need to log at the info level on a production service.


marktoda (approved):

marktoda:
some of the info logs are useful for debugging tho :thinking:

mikeki:
> some of the info logs are useful for debugging tho

Yeah, I agree, although we should probably choose those logs and mark as warn?

I'm wondering if there's any latency impact out of logging, but I also suspect that the library is probably supposed to make that faster?

What kind of debugging do you recall doing with these logs?

mikeki:
We will merge this branch to collect data and see if the log level is having any impact in latency, it will be reverted soon after.

PR #181, opened 2023-05-08
feat: setup dynamodb for unit tests in routing-api repo

lynnshaoyu:
1. Set up dynamo db local for use in unit tests
2. Creates an example "proof of concept" test: routing-api/test/unit/dynamoDbExampleTest.test.ts


test/mocha/dynamoDBLocalFixture.ts
```diff
@@ -0,0 +1,52 @@
...
+
+process.env.AWS_ACCESS_KEY_ID = 'my_access_key'
+process.env.AWS_SECRET_ACCESS_KEY = 'my_secret_key'
+
+const dbPort = Number(process.env.DYNAMODB_LOCAL_PORT || 8000)
+
+let dbInstance: ChildProcess | undefined
+;(global as any)['__DYNAMODB_LOCAL__'] = true
+
+export const mochaGlobalSetup = async () => {
```
mikeki, line 14: I think we should wrap this all around a Try catch block to return a good error message saying that the DynamoDB Local was not instantiated (This is where things were failing when your Java version wasn't working, but it was not clear)

test/unit/dynamoDbExampleTest.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const UNI_MAINNET = new Token(
+  ChainId.MAINNET,
+  '0x1f9840a85d5af5bf1d1762f925bdaddc4201f984',
+  18,
+  'UNI',
+  'Uniswap'
+)
+
+describe('DynamoRouteCache', async () => {
```
mikeki, line 41: I would place this test in the location of the DynamoRouteCaching file, so this becomes the actual test for that file

lynnshaoyu: done

test/mocha/dynamoDBLocalFixture.ts
lynnshaoyu: ooh sg done

mikeki (approved):

PR #184, opened 2023-05-11
CI: add gated build against routing-api via GitHub Actions

jsy1218:
## What?
[routing-api](https://github.com/Uniswap/routing-api) doesn't have gated build as of today. Regressions can be merged into the default branch. We want to incrementally improve the due diligence by first introducing the gated-build.

## Why?
With limited engineering resources, it's prudent we apply as much automated gating as possible, so that we reduce the feedback cycle on catching the regressions.

## How?
First step is to enable the gated build with the unit tests. It gets auto-triggered upon a new push as well as a pull request.

## Testing?
All the github action checks have passed.

## Screenshots (optional)
N/A

## Anything Else?
Once the gated build workflow works and gets merged into the main branch, we should update the merge policy to make it required as well.


mikeki:
It doesn't look like it is working. You could also check how it's done in Smart Order Router: https://github.com/Uniswap/smart-order-router/blob/main/.github/workflows/tests.yml in case that helps

jsy1218:
> It doesn't look like it is working. You could also check how it's done in Smart Order Router: https://github.com/Uniswap/smart-order-router/blob/main/.github/workflows/tests.yml in case that helps

Thanks. I think the root cause is access issue, based on the console output from [Lint](https://github.com/Uniswap/routing-api/actions/runs/4952970171/jobs/8859940909?pr=184). Ironically the Gated [Build](https://github.com/Uniswap/routing-api/actions/runs/4952970172/jobs/8859940925?pr=184) succeeded. Will loop deeper into the access issue.

jsy1218:
> How do we make the Unit Test Required?
> 
> Screenshot 2023-05-17 at 9 58 13 AM

My understanding is that the branch protection rule can be [configured](https://github.com/Uniswap/routing-api/settings/branch_protection_rules/29014700) to include the unit-tests once it gets merged. Right now the rule does not detect unit-tests since it's not on main yet:

Screenshot 2023-05-17 at 12 03 34 PM

Hence I want to try to merge this PR into main first, and see if I can update the rule.

mikeki (approved):

PR #190, opened 2023-05-15
modify linter to use nvm

jsy1218:
## What?
[routing-api](https://github.com/Uniswap/routing-api) contains nvm after the linter workflow. Linter is hardcoded to use node 16, but we want to make it consistent with the nvm managed version.

## Why?
We want to minimize the node version discrepancy across the [routing-api](https://github.com/Uniswap/routing-api) project.

## How?
Start with [routing-api](https://github.com/Uniswap/routing-api), instead of having the usual hard-coded node version(s) to be installed, we use nvm to install the node.

## Testing?
We will check whether the linter workflow passes. 

## Screenshots (optional)
N/A

## Anything Else?
At some point. we should check all the node projects to 1) patch with nvm, if missing 2) update the workflow(s) to use nvm instead.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,11 +15,9 @@ jobs:
       - name: Check out Git repository
         uses: actions/checkout@v2
 
-      - name: Set up node
-        uses: actions/setup-node@v1
-        with:
-          node-version: 16.x
-          registry-url: https://registry.npmjs.org
+      - name: Set the correct Node version using nvm
+        shell: bash -l {0} 
+        run: nvm install
 
       - name: Install dependencies
         run: npm ci


mikeki (approved):

.github/workflows/lint.yml
```diff
@@ -15,19 +15,17 @@ jobs:
...
       - name: Run linters
-        uses: wearerequired/lint-action@v1
+        uses: wearerequired/lint-action@v2
         with:
           github_token: ${{ secrets.github_token }}
           prettier: true
-          auto_fix: true
-          prettier_extensions: 'css,html,js,json,jsx,md,sass,scss,ts,tsx,vue,yaml,yml,sol'
+          auto_fix: false
+          prettier_extensions: 'css,html,js,json,jsx,md,sass,scss,ts,tsx,vue,yaml,yml,sol'
```
zzmp, line 31: I strongly recommend you replace lint-action with your own `yarn lint`, as this has better versioning control, will better respect any lint configuration you set up, and will be more testable locally.

.github/workflows/lint.yml
```diff
@@ -15,19 +15,17 @@ jobs:
       - name: Check out Git repository
         uses: actions/checkout@v2
 
-      - name: Set up node
-        uses: actions/setup-node@v1
-        with:
-          node-version: 16.x
-          registry-url: https://registry.npmjs.org
+      - name: Set the correct Node version using nvm
+        shell: bash -l {0}
```
zzmp, line 19: What does this mean? Why not just set `bash`?

.github/workflows/lint.yml
```diff
@@ -15,19 +15,17 @@ jobs:
...
+      - name: Set the correct Node version using nvm
+        shell: bash -l {0} 
+        run: nvm install
 
       - name: Install dependencies
         run: npm ci
 
       - name: Run linters
-        uses: wearerequired/lint-action@v1
+        uses: wearerequired/lint-action@v2
```
zzmp, line 26: if you're changing this, it's safer to use a git commit than a major version, as the minor version can be bumped to change the package maliciously.

.github/workflows/lint.yml
zzmp, line 31: Instead of adding a .prettierignore, why not just remove yaml and yml from the list of prettier_extensions?

jsy1218: For sure, I also feel concerned to see the non-reproducibility on local.

I actually saw smart-order-route uses npm [lint](https://github.com/Uniswap/smart-order-router/blob/main/.github/workflows/lint.yml), and curious to see your thoughts. Do you think we should strive to use `yarn` or `npm` with `nvm`?

.github/workflows/lint.yml
jsy1218: I didn't spend much thought into this one tbh. I looked at this [article](https://snippets.khromov.se/use-nvm-in-github-actions-when-building-a-node-js-project/), and decided to use it verbatim. For the reproducibility, I didn't opt into the github action nvm from the [marketplace](https://github.com/marketplace/actions/set-up-node-using-nvm).

Will try using `bash` only and see if it works.

jsy1218: It appears only setting bash will fail, see failed workflow [run](https://github.com/Uniswap/routing-api/actions/runs/4995256903/jobs/8946899274?pr=190) vs successful workflow [run](https://github.com/Uniswap/routing-api/actions/runs/4994295293/jobs/8944653089). 

I noticed in the failed run:
```
nvm install
  shell: /usr/bin/bash --noprofile --norc -e -o pipefail {0}
/home/runner/work/_temp/9e1ec01[4](https://github.com/Uniswap/routing-api/actions/runs/4995256903/jobs/8946899274?pr=190#step:3:5)-ee21-422d-a00b-8f226d4a076d.sh: line 1: nvm: command not found
Error: Process completed with exit code 127.
```

in the successful run:
```
nvm install
  shell: /usr/bin/bash -l {0}
Found '/home/runner/work/routing-api/routing-api/.nvmrc' with version <18.1[3](https://github.com/Uniswap/routing-api/actions/runs/4994295293/jobs/8944653089#step:3:3).0>
Downloading and installing node v18.13.0...

```

zzmp (approved):

.github/workflows/lint.yml
zzmp: it won't affect package.json or package-lock.json, but pacakge.json/package-lock.json give you stronger guarantees than the version here. you could lock it to a more explicit version by specifying a git commit hash from the wearerequired/lint-action repo.

you don't need to change this now, though.

.github/workflows/lint.yml
zzmp: the web team uses `yarn`, not `npm`. we also use `nvm`, but not in our github actions workflow files.

PR #195, opened 2023-05-22
feat: sample buckets 3 for usdc to any token dashboard 

lynnshaoyu:


mikeki (changes requested):
Having that many buckets is a little bit counterproductive, we have 22~35% of requests for amounts lower than 750, if we divide them in this many buckets we will increase the number of cache misses.

lib/handlers/router-entities/route-caching/cached-routes-configuration.ts
```diff
@@ -159,9 +159,12 @@ export const CACHED_ROUTES_CONFIGURATION: Map<string, CachedRoutesStrategy> = ne
       tradeType: TradeType.EXACT_INPUT,
       chainId: ChainId.MAINNET,
       buckets: [
-        new CachedRoutesBucket({ bucket: 100, cacheMode: CacheMode.Tapcompare, withLastNCachedRoutes: 10 }),
```
mikeki, line 162: Let's instead try to just Darkmode this bucket for 100 and see how things change

mikeki (approved):

jsy1218:
There's a test failure, seems like test update needed

jsy1218:
> Having that many buckets is a little bit counterproductive, we have 22~35% of requests for amounts lower than 750, if we divide them in this many buckets we will increase the number of cache misses.

I wonder if we have a way to canary changeset like this?

PR #200, opened 2023-05-25
Bump SOR to 3.11.4

mikeki:


cgkol (approved):

jsy1218:
Slightly confused, with caret(`^`), wouldn't the latest minor or patch version of `smart-order-route` library gets picked up? In other words, if we want to be super strict on bumping up the `smart-order-route` version, shall we remove `^`?

jsy1218 (approved):

PR #201, opened 2023-05-25
[feat] [ROUTE-9] v3 dynamo cache pool provider

jsy1218:
## What?
In routing-api, we create a [CachingV3PoolProvider](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/injector-sor.ts#L242) which is defined in [SOR](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/v3/caching-pool-provider.ts#L20).

We can create a new DynamoDBCachingV3PoolProvider in the routing-api source code that implements IV3PoolProvider, and uses a DynamoDB Table for storing the Pools into cache.

It is important to use the block number as a sort key for the cache to make sure that we are storing the information of the pools at a given block height, and for consistency of quotes.
 
## Why?
We see misquotes across v2 and v3. We think the existing in-memory cache is one source of the issue.

## How?
We will begin replacing v3 in-memory cache with distributed cache. 

## Testing?
All unit tests and integ tests passed.

## Screenshots (optional)
N/A

## Anything Else?
BlockNumber are passed into [V2Quoter](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/quoters/v2-quoter.ts) and [MixedQuoter](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/quoters/mixed-quoter.ts), but not [V3Quoter](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/quoters/v3-quoter.ts) . I think we would want to remain as is for now, and test the [MixedQuoter](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/quoters/mixed-quoter.ts) after dynamo caching V3.


lib/handlers/router-entities/pool-caching/v3/dynamo-pool-provider.ts
```diff
@@ -0,0 +1,110 @@
+import { ChainId, ICache, IV3PoolProvider, V3PoolAccessor } from '@uniswap/smart-order-router'
+import { Token } from '@uniswap/sdk-core'
+import { ProviderConfig } from '@uniswap/smart-order-router/build/main/providers/provider'
+import { FeeAmount, Pool } from '@uniswap/v3-sdk'
+import { DynamoDB } from 'aws-sdk'
+
+export class DynamoPoolProvider implements IV3PoolProvider {
```
jsy1218, line 7: @mikeki I find it a bit awkward that I'm copying and pasting a lot of code from [caching-pool-provider.ts](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/v3/caching-pool-provider.ts), but I think it's better to minimize the refactoring in [smart-order-router](https://github.com/Uniswap/smart-order-router) right? Also in this case, we will be able to delete [caching-pool-provider.ts](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/v3/caching-pool-provider.ts) upon migration to dynamo cache.

lib/handlers/router-entities/pool-caching/v3/dynamo-pool-provider.ts
```diff
@@ -0,0 +1,109 @@
...
+  private readonly ddbClient: DynamoDB.DocumentClient
+  private readonly tableName: string
+  private readonly ttlMinutes: number
+  private POOL_CACHE_KEY = (chainId: ChainId, address: string) => `pool-${chainId}-${address}`
+
+  constructor(protected chainId: ChainId, protected poolProvider: IV3PoolProvider) {
```
mikeki, line 13: you might need to receive the table name here, roght?

lib/handlers/router-entities/pool-caching/v3/dynamo-pool-provider.ts
mikeki: I think you no longer are doing alot of copy pasting?

lib/handlers/router-entities/pool-caching/v3/dynamo-pool-provider.ts
```diff
@@ -0,0 +1,109 @@
...
+        if (pool) {
+          poolAddressToPool[address] = pool
+          // TODO: marshall the pool object
+          const binaryCachedPool: Buffer = Buffer.from(JSON.stringify(pool))
+
+          const putParams = {
+            TableName: this.tableName,
+            Item: {
+              poolAddress: this.POOL_CACHE_KEY(this.chainId, address),
+              blockNumber: (await providerConfig?.blockNumber) ?? 0,
```
mikeki, line 89: maybe we should attempt to guarantee that the block number is there, if it's not then it doesn't make sense to cache it, right?

lib/handlers/router-entities/pool-caching/v3/dynamo-pool-provider.ts
```diff
@@ -0,0 +1,109 @@
...
+              item: binaryCachedPool,
+              ttl: this.ttlMinutes,
```
mikeki, line 91: I think Dynamo's ttl works different than this, it is based on a timestamp, so you can have a ttl in the class that defines how many seconds after the current timestamp (check the DynamoRouteCachingProvider)

lib/handlers/router-entities/pool-caching/v3/dynamo-pool-provider.ts
jsy1218: Yes from [cache-dynamo-pool.ts](https://github.com/Uniswap/routing-api/pull/201/files#diff-5339f16f1f7af49f5b71baf15da11c9f3f02fb225346eea97ececdf1c3fb42b2)

lib/handlers/router-entities/pool-caching/v3/dynamo-pool-provider.ts
jsy1218: Yes from [cache-dynamo-pool.ts](https://github.com/Uniswap/routing-api/pull/201/files#diff-5339f16f1f7af49f5b71baf15da11c9f3f02fb225346eea97ececdf1c3fb42b2)

lib/handlers/router-entities/pool-caching/v3/dynamo-pool-provider.ts
jsy1218: Yes from [cache-dynamo-pool.ts](https://github.com/Uniswap/routing-api/pull/201/files#diff-5339f16f1f7af49f5b71baf15da11c9f3f02fb225346eea97ececdf1c3fb42b2)

jsy1218:
## What?
In routing-api, we create a [CachingV3PoolProvider](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/injector-sor.ts#L242) which is defined in [SOR](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/v3/caching-pool-provider.ts#L20).

We can create a new DynamoDBCachingV3PoolProvider in the routing-api source code that implements IV3PoolProvider, and uses a DynamoDB Table for storing the Pools into cache.

It is important to use the block number as a sort key for the cache to make sure that we are storing the information of the pools at a given block height, and for consistency of quotes.
 
## Why?
We see misquotes across v2 and v3. We think the existing in-memory cache is one source of the issue.

## How?
We will begin replacing v3 in-memory cache with distributed cache. 

## Testing?
All unit tests and integ tests passed.

## Screenshots (optional)
N/A

## Anything Else?
BlockNumber are passed into [V2Quoter](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/quoters/v2-quoter.ts) and [MixedQuoter](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/quoters/mixed-quoter.ts), but not [V3Quoter](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/quoters/v3-quoter.ts) . I think we would want to remain as is for now, and test the [MixedQuoter](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/quoters/mixed-quoter.ts) after dynamo caching V3.

lib/handlers/router-entities/pool-caching/v3/cache-dynamo-pool.ts
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  override async has(partitionKey: string, sortKey?: number): Promise<boolean> {
+    return Promise.resolve((await this.get(partitionKey, sortKey)) !== undefined)
+  }
+
+  override async set(value: Pool, partitionKey: string, sortKey?: number): Promise<boolean> {
+    if (sortKey) {
+      // TODO: marshall the pool object
+      const binaryCachedPool: Buffer = Buffer.from(JSON.stringify(value))
```
mikeki, line 51: Can we guarantee that converting the `Pool` to JSON and from JSON will work correctly?

For CachedRoutes I had to create marshallers because I couldnt guarantee that the objects would get created back with a simple JSON.stringify and JSON.parse

lib/handlers/router-entities/cache-dynamo.ts
```diff
@@ -0,0 +1,38 @@
+import { DynamoDB } from 'aws-sdk'
+
```
mikeki, line 2: I would create a new folder for this, maybe? I feel like the root of `router-entities` is getting too crowded

lib/handlers/router-entities/pool-caching/v3/cache-dynamo-pool.ts
```diff
@@ -0,0 +1,78 @@
...
+    super({ tableName })
+  }
+
+  override async get(partitionKey: string, sortKey?: number): Promise<Pool | undefined> {
+    if (sortKey) {
+      const getParams = {
+        TableName: this.tableName,
+        Key: {
+          poolAddress: partitionKey,
+          blockNumber: sortKey,
```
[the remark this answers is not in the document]
jsy1218: aw what's the use case for fetching some past blocks? Is it when we don't know what's the latest the block number?

lib/handlers/router-entities/pool-caching/v3/cache-dynamo-pool.ts
```diff
@@ -0,0 +1,78 @@
...
+      } else {
+        return undefined
+      }
+    } else {
+      return undefined
+    }
+  }
+
+  override async has(partitionKey: string, sortKey?: number): Promise<boolean> {
+    return Promise.resolve((await this.get(partitionKey, sortKey)) !== undefined)
```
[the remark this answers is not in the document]
jsy1218: I can see your concern. We can remove `has` method

lib/handlers/router-entities/pool-caching/v3/cache-dynamo-pool.ts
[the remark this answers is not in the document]
mikeki: In case we want to optimistically use the previous block data, but I think we can go with what we have right now, and then see how it performs before making other changes :)

lib/handlers/router-entities/pool-caching/v3/cache-dynamo-pool.ts
jsy1218: That's a fair concern. We can use marshalling. Less chance with breaking forward compatible schema changes.

lib/handlers/marshalling/pool-marshaller.ts
```diff
@@ -1,6 +1,6 @@
 import { Pool } from '@uniswap/v3-sdk'
 import { FeeAmount } from '@uniswap/v3-sdk/dist/constants'
-import { MarshalledToken, TokenMarshaller } from './token-marshaller'
+import { MarshalledToken, TokenMarshaller } from '../router-entities/route-caching/marshalling/token-marshaller'
```
mikeki, line 3: it seems like this file was moved to a different place, let's move all of the marshallers, or none of them it feels confusing to have them separated.

mikeki (approved):
It looks good overall, let's just address the last comment I added, but this looks great!

PR #205, opened 2023-05-30
Bump SOR to 3.11.5

mikeki:
This version of SOR implements some changes with how we resolve tokens that were previously impossible to resolve (due to symbol returning bytes32 instead of string)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,7 +27,7 @@
         "@uniswap/permit2-sdk": "^1.2.0",
         "@uniswap/router-sdk": "^1.3.0",
         "@uniswap/sdk-core": "^3.0.1",
-        "@uniswap/smart-order-router": "^3.11.4",
+        "@uniswap/smart-order-router": "^3.11.5",
         "@uniswap/token-lists": "^1.0.0-beta.24",
         "@uniswap/universal-router-sdk": "^1.3.0",
         "@uniswap/v2-sdk": "^3.0.0",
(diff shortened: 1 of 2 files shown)


cgkol (approved):
did we test locally that this wont break it?

jsy1218:
> did we test locally that this wont break it?

My understanding is that the E2E changes like this (routing-api & smart-order-route (including ABI) & contracts) are hard to get canary tested. Testing in local won't catch this regression, because it involves routing-api & smart-order-route calling the smart contracts via the ABI with particular token pairs, and we only have basic integ tests in our routing-api test suite. @mikeki Feel free to correct me.

jsy1218 (approved):

mikeki:
I did a test on SOR and the test passed, so at least when it comes to tests this is looking good

cgkol:
> > did we test locally that this wont break it?
> 
> My understanding is that the E2E changes like this (routing-api & smart-order-route (including ABI) & contracts) are hard to get canary tested. Testing in local won't catch this regression, because it involves routing-api & smart-order-route calling the smart contracts via the ABI with particular token pairs, and we only have basic integ tests in our routing-api test suite. @mikeki Feel free to correct me.

fwiw, @mikeki @jsy1218, siyu and I talked on Slack, and what I meant was that we could do this on our personal account to test the FE flow. maybe an overkill, but if its already presetup we can check some normal flows for quoting to make sure nothing breaks. Again, might be an overkill for this PR but does not hurt to test when we bump this in general, regardless of how big the change is

PR #208, opened 2023-06-01
feat: turn on cache for usdc/any token routes up to 1000 usdc input

lynnshaoyu:


mikeki (approved):

jsy1218 (approved):

jsy1218:
Second thought on enabling the live traffic for USDC - can we raise the bar by e2e testing via personal aws account? I don't know if we need to swap the API in the interface for this change, or can we just curl the CFN API from cmd?

mikeki (approved):

mikeki:
> Second thought on enabling the live traffic for USDC - can we raise the bar by e2e testing via personal aws account? I don't know if we need to swap the API in the interface for this change, or can we just curl the CFN API from cmd?

There's really not much we get out of testing this change in the personal account other than verifying it deploys, because it requires traffic to actually work, so for this specific change I don't think that's necessary

PR #209, opened 2023-06-01
[ROUTE-9] Weave the v3 pools cache dynamo table into the lambda stack

jsy1218:
## What?
After [PR](https://github.com/Uniswap/routing-api/pull/201), we should now provision the v3 dynamo table. 
 
## Why?
Because we need a new table for lib/handlers/pools/pool-caching/v3/dynamo-caching-pool-provider.ts to interact with.

## How?
We do so by adding the dynamo DB stack into the lambda stack. It should auto-populate the new table.

## Testing?

- All unit tests and integ tests passed.
- cdk deploy via personal aws account succeeded. new V3PoolDB table got provisioned

## Screenshots (optional)
Screenshot 2023-06-02 at 10 50 27 AM
Screenshot 2023-06-05 at 1 45 44 PM

## Anything Else?
Once we confirm we provision the table, next step we can implement the dark traffic into the dynamo cache, and compare the quote accuracy.


bin/stacks/routing-database-stack.ts
```diff
@@ -5,8 +5,22 @@ import { Construct } from 'constructs'
 
 export interface RoutingDatabaseStackProps extends cdk.NestedStackProps {}
 
+export const V3PoolsDynamoDbTableName = 'V3PoolsCachingDB'
+export const V3PoolsDynamoDbPartitionKeyName = 'poolAddress'
+export const V3PoolsDynamoDbSortKeyName = 'blockNumber'
+export const TTLAttributeName = 'ttl'
+
+export const V3PoolsDynamoDbProps: aws_dynamodb.TableProps = {
```
mikeki, line 13: I would prefer that this configuration is inline with the Dynamo Table creation, it feels like a layer of indirection for it to be here.

bin/stacks/routing-database-stack.ts
```diff
@@ -5,8 +5,22 @@ import { Construct } from 'constructs'
 
 export interface RoutingDatabaseStackProps extends cdk.NestedStackProps {}
 
+export const V3PoolsDynamoDbTableName = 'V3PoolsCachingDB'
```
mikeki, line 8: For consistency, could we do the same in the CachedRoutes Table?

mikeki (commented):
This will not yet populate the table, it is just creating the AWS instance, right?

bin/stacks/routing-database-stack.ts
jsy1218: > This will not yet populate the table, it is just creating the AWS instance, right?

We clarified offline. "populate the table" = provision the table, but I agree populate sounds more like writing entries into the table.

bin/stacks/routing-database-stack.ts
jsy1218: Yep. I did logical grouping of DynamoDB constants and refactored cached routes table attributes accordingly.

jsy1218:
> This will not yet populate the table, it is just creating the AWS instance, right?

We clarified offline. "populate the table" = provision the table, but I agree populate sounds more like writing entries into the table.

lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
```diff
@@ -224,8 +225,8 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
       const putParams = {
         TableName: this.tableName,
         Item: {
-          pairTradeTypeChainId: partitionKey.toString(),
-          protocolsBucketBlockNumber: sortKey.fullKey(),
+          [DynamoDBTableProps.CacheRouteDynamoDbTable.PartitionKeyName]: partitionKey.toString(),
+          [DynamoDBTableProps.CacheRouteDynamoDbTable.SortKeyName]: sortKey.fullKey(),
           item: binaryCachedRoutes,
           ttl: ttl,
```
mikeki, line 231: Should we do the same here with the ttl attribute?

jsy1218: yep

mikeki (approved):

PR #210, opened 2023-06-01
[ROUTE-9] Reduce the v3 pool TTL to 1 minute

jsy1218:
We decided to reduce the TTL to 1 minute to begin with. The average block settlement time is between 10-15 seconds. We can adjust down as appropriate once we launch the distributed pools cache.


jsy1218:
> Are there any threads/discussion/docs as to how we decide about this? If not can we summarize here why?

@mikeki and I had a brief chat. 1min is a comfortable heuristics, and we can adjust down as needed.

mikeki:
@cgkol We really only need the TTL to be larger than 13 sec, because we aren't looking at the entries past the current block, we could change this TTL if we ever decide to change this behavior.

cgkol:
@mikeki @jsy1218 i was just thinking, is 1 minute too long given that the data can be up to 5 blocks stale? Also please add my name on future comments because otherwise i dont get the notif :')

mikeki:
@cgkol well, we dont really use the data after 1minute, thats just controlling the DB growth but the sort key makes sure that we only fetch the entry for the latest block

cgkol:
@mikeki i see i see, makes sense

cgkol (approved):

PR #211, opened 2023-06-01
[feat] [ROUTE-9] Add traffic switch between in-memory cache pool provider and dynamo distributed cache pool provider for v3

jsy1218:
## What?
After [PR](https://github.com/Uniswap/routing-api/pull/209), we are ready to switch the traffic to the dynamo caching incrementally. In switching, we also want to evaluate the quality of dynamo caching, so that we will want to get the accuracy of the pool pricing for both current pool provider (in-memory cache) and the target pool provider (dynamo cache). 
 
## Why?
[caching-pool-provider.ts](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/v3/caching-pool-provider.ts) is on the critical path of [v3-quoter.ts](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/quoters/v3-quoter.ts) and [mixed-quoter.ts](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/quoters/mixed-quoter.ts). We don't want to blatantly switch the traffic without knowing if we improve/break the quote path in routing-api. Also right now we don't have any visibility into the [caching-pool-provider.ts](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/v3/caching-pool-provider.ts) pool pricing accuracy, and thus we need datapoints to back up our claim that [dynamo-caching-pool-provider.ts
](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/pools/pool-caching/v3/dynamo-caching-pool-provider.ts) will improve both accuracy and tail latency.

## How?
We introduce [traffic-switch-pool-provider.ts](https://github.com/Uniswap/routing-api/pull/211/files#diff-c48ee34e6d3121ea37c21a8689474e0bfa3d6481c9ce878b85b10d766571dcbb) that will do two things:

1. Sample the request to get the accuracy. The sampling rate is set at 1%. For each sampling, we will get the pool pricing accuracy for the existing in-memory cache as well as the dynamo cache, comparing against the actual call to the RPC provider.
2. Traffic switch between the existing in-memory cache and the dynamo cache. We start at 0%, because we don't have enough datapoints to switch even 1%.

## Testing?
Unit tests and integ-tests passed against my cdk stack instance. Also created a [dashboard](https://cloudwatch.amazonaws.com/dashboard.html?dashboard=RoutingPoolProviderDashboard&context=eyJSIjoidXMtZWFzdC0xIiwiRCI6ImN3LWRiLTU3NTM3MjQxMjAzMSIsIlUiOiJ1cy1lYXN0LTFfd2ROSFc1d2Q2IiwiQyI6IjQ2YmdscTNybTJvNjBhYWVlcGpwMGtlZW02IiwiSSI6InVzLWVhc3QtMTowMzNmZWM3Ni00YzkyLTQ2YjctOGY4Ni1kZDYzNjVjZDgyOGIiLCJPIjoiYXJuOmF3czppYW06OjU3NTM3MjQxMjAzMTpyb2xlL3NlcnZpY2Utcm9sZS9DV0RCU2hhcmluZy1SZWFkT25seUFjY2Vzcy1VMUUyWjJOMSIsIk0iOiJVc3JQd1NpbmdsZSJ9) to track the accuracy and traffic switch.

## Screenshots (optional)
Screenshot 2023-06-07 at 3 40 32 PM


lib/handlers/injector-sor.ts
```diff
@@ -240,11 +242,23 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
             break
         }
 
-        const v3PoolProvider = new CachingV3PoolProvider(
+        const sourceOfTruthV3PoolProvider = new V3PoolProvider(chainId, multicall2Provider)
```
mikeki, line 245: maybe rename as `underlyingV3PoolProvider` or `v3PoolProvider` or `onchainV3PoolProvider`

lib/handlers/injector-sor.ts
```diff
@@ -240,11 +242,23 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
...
+        const currentV3PoolProvider = new CachingV3PoolProvider(
           chainId,
-          new V3PoolProvider(chainId, multicall2Provider),
+          sourceOfTruthV3PoolProvider,
           new NodeJSCache(new NodeCache({ stdTTL: 180, useClones: false }))
         )
+        const targetV3PoolProvider = new DynamoDBCachingV3PoolProvider(
```
mikeki, line 251: let's also rename this as `dynamoV3PoolProvider` or `awsCachingV3PoolProvider` or something like that

lib/handlers/pools/provider-migration/v3/traffic-switch-pool-provider.ts
```diff
@@ -0,0 +1,103 @@
...
+  } {
+    // The underlying logic for getting the pool address is always from the
+    // source of truth pool provider, and the result is deterministic.
+    // It doesn't matter which pool provider we use, so we can start with the target
+    // pool provider, which will delegate to the source of truth pool provider.
+    return this.targetPoolProvider.getPoolAddress(tokenA, tokenB, feeAmount)
+  }
+
+  async getPools(tokenPairs: [Token, Token, FeeAmount][], providerConfig?: ProviderConfig): Promise<V3PoolAccessor> {
+    const currentProviderPools = await this.currentPoolProvider.getPools(tokenPairs, providerConfig)
```
mikeki, line 46: I think we should avoid calling them when not necessary, if we aren't sampling nor switching then we don't need to fetch the `targetProviderPools`, same goes for when we are switching.

lib/handlers/pools/provider-migration/v3/traffic-switch-pool-provider.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+  }
+
+  private async sampleTraffic(
+    tokenPairs: [Token, Token, FeeAmount][],
+    currentProviderPools: V3PoolAccessor,
+    targetProviderPools: V3PoolAccessor,
+    providerConfig?: ProviderConfig
+  ): Promise<void> {
+    const truthProviderPools = await this.sourceOfTruthPoolProvider.getPools(tokenPairs, providerConfig)
```
mikeki, line 69: I get what we are doing here, comparing the old and new methods against the actual sourceOfTruth... this could slow down the sampled requests, right?

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  async getPools(tokenPairs: [Token, Token, FeeAmount][], providerConfig?: ProviderConfig): Promise<V3PoolAccessor> {
+    const currentProviderPools = await this.currentPoolProvider.getPools(tokenPairs, providerConfig)
+
+    metric.putMetric('V3_POOL_PROVIDER_POOL_TRAFFIC_TOTAL', 1, MetricLoggerUnit.None)
+    const sampleTraffic = this.SHOULD_SAMPLE_TRAFFIC()
+    if (sampleTraffic) {
+      metric.putMetric('V3_POOL_PROVIDER_POOL_TRAFFIC_SAMPLING', 1, MetricLoggerUnit.None)
+      const targetProviderPools = await this.targetPoolProvider.getPools(tokenPairs, providerConfig)
```
mikeki, line 52: Shouldn't this be gated behind an `if (!switchTraffic)`?

jsy1218: The idea is that we want to shadow sample 1% of the request, meanwhile we do not switch the traffic yet. We don't know yet how accurate the dynamo caching provider will be.

cgkol:
@jsy1218 my bad i accidentally clicked approved i was still reviewing, give me a few; the description comment still stands tho!

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
```diff
@@ -0,0 +1,120 @@
...
+
+      // If we need to sample the traffic, we don't want to make it a blocking I/O
```
cgkol, line 54: if we want to swtich traffic, then this SHOULD be a blocking call right?

cgkol: i guess you are awaiting at line 64 but you are calculating the same thing twice it seems?

lib/handlers/pools/provider-migration/v3/pool-provider-traffic-switch-configuration.ts
```diff
@@ -0,0 +1,4 @@
+export const POOL_PROVIDER_TRAFFIC_SWITCH_CONFIGURATION = {
+  switchPercentage: 0.0,
```
cgkol, line 2: these should proabbly be secrets/flags so we can unroll them faster if we accidentally push something we do not want

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
```diff
@@ -0,0 +1,120 @@
...
+    currentProviderPools: V3PoolAccessor,
+    targetProviderPools: V3PoolAccessor,
+    providerConfig?: ProviderConfig
+  ): Promise<void> {
+    const truthProviderPools = await this.sourceOfTruthPoolProvider.getPools(tokenPairs, providerConfig)
+
+    truthProviderPools.getAllPools().forEach((pool: Pool) => {
+      const currentProviderPool = currentProviderPools.getPool(pool.token0, pool.token1, pool.fee)
+      if (!currentProviderPool) {
+        // We don't expect missing pool, but export metric in case we see any
```
cgkol, line 79: this can happen if there is a cache miss still right?

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
```diff
@@ -0,0 +1,120 @@
...
+        metric.putMetric('V3_POOL_PROVIDER_POOL_CURRENT_MISSING', 1, MetricLoggerUnit.None)
+        log.info(
+          `v3 Pool ${pool.token0.symbol} ${pool.token1.symbol} ${pool.fee} not found in the current pool provider.`
+        )
+      } else {
+        const sameQuote = currentProviderPool.token0Price.scalar.equalTo(pool.token0Price.scalar)
```
cgkol, line 85: we should probably add some jitter here right? exactly the same quote is a tall ask

cgkol: same for the target one

cgkol (commented):
looks good, added some comments!

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
mikeki: we will accomplish this as long as we have 0% in the switchTraffic config, right? let me propose some code changes here

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
```diff
@@ -0,0 +1,120 @@
...
+      this.sampleTraffic(tokenPairs, currentProviderPools, targetProviderPools, providerConfig)
+    }
+
+    const switchTraffic = this.SHOULD_SWITCH_TRAFFIC()
+    if (!switchTraffic) {
+      metric.putMetric('V3_POOL_PROVIDER_POOL_TRAFFIC_CURRENT', 1, MetricLoggerUnit.None)
+      return currentProviderPools
+    } else {
+      metric.putMetric('V3_POOL_PROVIDER_POOL_TRAFFIC_TARGET', 1, MetricLoggerUnit.None)
+      return await this.targetPoolProvider.getPools(tokenPairs, providerConfig)
```
mikeki, line 64: I think this should be:

```
    const sampleTraffic = this.SHOULD_SAMPLE_TRAFFIC()
    const switchTraffic = this.SHOULD_SWITCH_TRAFFIC()
    let currentProviderPools
    let targetProviderPools
    
    metric.putMetric('V3_POOL_PROVIDER_POOL_TRAFFIC_TOTAL', 1, MetricLoggerUnit.None)
    
    if (sampleTraffic) {
      metric.putMetric('V3_POOL_PROVIDER_POOL_TRAFFIC_SAMPLING', 1, MetricLoggerUnit.None)
      
      currentProviderPools = await this.currentPoolProvider.getPools(tokenPairs, providerConfig)
      targetProviderPools = await this.targetPoolProvider.getPools(tokenPairs, providerConfig)
       // If we need to sample the traffic, we don't want to make it a blocking I/O
      this.sampleTraffic(tokenPairs, currentProviderPools, targetProviderPools, providerConfig)
    }
    
    if (switchTraffic) {
      metric.putMetric('V3_POOL_PROVIDER_POOL_TRAFFIC_TARGET', 1, MetricLoggerUnit.None)
      
      return targetProviderPools ?? await this.targetPoolProvider.getPools(tokenPairs, providerConfig)
    } else {
      metric.putMetric('V3_POOL_PROVIDER_POOL_TRAFFIC_CURRENT', 1, MetricLoggerUnit.None)
      
      return currentProviderPools ?? await this.currentPoolProvider.getPools(tokenPairs, providerConfig)
    }
```

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
jsy1218: I agree that we might need to jitter depending on which pool, based on the cached routes experience from @mikeki . Had a quick chat with @mikeki , and we think we want to ship it to prod to see how the accuracy looks like this week.

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
jsy1218: Actually it shouldn't. The reason is because if there's a cache miss from `CachingV3PoolProvider`, then it will delegate to the actual provider `V3PoolProvider` to get the pool.

lib/handlers/pools/provider-migration/v3/pool-provider-traffic-switch-configuration.ts
jsy1218: I agree, we can probably fast follow once we ship something this week. Since the traffic is at 0% percent, we can fast follow next week if deemed needed.

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
```diff
@@ -0,0 +1,120 @@
...
+    feeAmount: FeeAmount
+  ): {
+    poolAddress: string
+    token0: Token
+    token1: Token
+  } {
+    // The underlying logic for getting the pool address is always from the
+    // source of truth pool provider, and the result is deterministic.
+    // It doesn't matter which pool provider we use, so we can start with the target
+    // pool provider, which will delegate to the source of truth pool provider.
```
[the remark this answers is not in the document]
jsy1218: So `TrafficSwitchV3PoolProvider.getPoolAddress(..)` calls `DynamoDBCachingV3PoolProvider.getPoolAddress(..)`, and `DynamoDBCachingV3PoolProvider.getPoolAddress(..)` calls `V3PoolProvider.getPoolAddress(..)`. 

Even if we use `CachingV3PoolProvider, the call chain will be:
`TrafficSwitchV3PoolProvider.getPoolAddress(..)` calls `CachingV3PoolProvider.getPoolAddress(..)`, and `CachingV3PoolProvider.getPoolAddress(..)` calls `V3PoolProvider.getPoolAddress(..)`

So the end result would be the same.

lib/handlers/pools/provider-migration/v3/pool-provider-traffic-switch-configuration.ts
cgkol: perfect, can we add a jira ticket for this so it does not get forgotten please?

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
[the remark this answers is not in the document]
cgkol: oh interesting, thank you for the detailed explanation. So `V3PoolProvider.getPoolAddress` is the `true` provider?

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
cgkol: sorry n00b question, but does the delegation happen within the `CachingV3PoolProvider`?

lib/handlers/pools/provider-migration/v3/pool-provider-traffic-switch-configuration.ts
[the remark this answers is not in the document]
jsy1218: agree, moved to the `lib/handlers/pools/util` folder

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
jsy1218: Modified the call structure, so that the call to currentProviderPools & targetProviderPools will be at most once respectively.

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
jsy1218: Nw, let me copy paste the code block:

```
export class CachingV3PoolProvider implements IV3PoolProvider {
  private POOL_KEY = (chainId: ChainId, address: string) =>
    `pool-${chainId}-${address}`;

  /**
   * Creates an instance of CachingV3PoolProvider.
   * @param chainId The chain id to use.
   * @param poolProvider The provider to use to get the pools when not in the cache.
   * @param cache Cache instance to hold cached pools.
   */
  constructor(
    protected chainId: ChainId,
    protected poolProvider: IV3PoolProvider,
    private cache: ICache<Pool>
  ) {}

  public async getPools(
    tokenPairs: [Token, Token, FeeAmount][],
    providerConfig?: ProviderConfig
  ): Promise<V3PoolAccessor> {
    const poolAddressSet: Set<string> = new Set<string>();
    const poolsToGetTokenPairs: Array<[Token, Token, FeeAmount]> = [];
    const poolsToGetAddresses: string[] = [];
    const poolAddressToPool: { [poolAddress: string]: Pool } = {};

    for (const [tokenA, tokenB, feeAmount] of tokenPairs) {
      const { poolAddress, token0, token1 } = this.getPoolAddress(
        tokenA,
        tokenB,
        feeAmount
      );

      if (poolAddressSet.has(poolAddress)) {
        continue;
      }

      poolAddressSet.add(poolAddress);

      const cachedPool = await this.cache.get(
        this.POOL_KEY(this.chainId, poolAddress)
      );
      if (cachedPool) {
        poolAddressToPool[poolAddress] = cachedPool;
        continue;
      }

      poolsToGetTokenPairs.push([token0, token1, feeAmount]);
      poolsToGetAddresses.push(poolAddress);
    }

    log.info(
      {
        poolsFound: _.map(
          Object.values(poolAddressToPool),
          (p) => `${p.token0.symbol} ${p.token1.symbol} ${p.fee}`
        ),
        poolsToGetTokenPairs: _.map(
          poolsToGetTokenPairs,
          (t) => `${t[0].symbol} ${t[1].symbol} ${t[2]}`
        ),
      },
      `Found ${
        Object.keys(poolAddressToPool).length
      } V3 pools already in local cache. About to get liquidity and slot0s for ${
        poolsToGetTokenPairs.length
      } pools.`
    );

    if (poolsToGetAddresses.length > 0) {
      const poolAccessor = await this.poolProvider.getPools(
        poolsToGetTokenPairs,
        providerConfig
      );
      for (const address of poolsToGetAddresses) {
        const pool = poolAccessor.getPoolByAddress(address);
        if (pool) {
          poolAddressToPool[address] = pool;
          await this.cache.set(this.POOL_KEY(this.chainId, address), pool);
        }
      }
    }
```

lib/handlers/pools/provider-migration/v3/pool-provider-traffic-switch-configuration.ts
jsy1218: @cgkol  ack, added ticket ROUTE-26 into the backlog.

lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts
jsy1218: Adopted this style. Indeed it minimizes the RPC calls.

mikeki (approved):

jsy1218:
Updated the screenshot of the [description](https://github.com/Uniswap/routing-api/pull/211#issue-1737229504) to show I `cdk deploy`ed again and ensured the metrics still getting exported after the suggested refactoring.

cgkol (approved):
yay

PR #215, opened 2023-06-06
chore: remove 2xx alarm for now

lynnshaoyu:
Removing for now due to too many non-actionable alerts. We also have alarms for 5xx and 4xx errors (% greater than a certain threshold) so removing this won't cause us to miss any of the errors we "actually care about".


jsy1218:
After we remove the programmatic alarm codepath, do we need to manually delete the alarm in the console? Or would it automatically get deleted? It's fine to check this after the codepipeline has released.

jsy1218 (approved):

PR #219, opened 2023-06-07
Add 5XX errors by chain metric

mikeki:
This PR makes a few changes:
1. Adds a 5XX error metric in the quote endpoint
2. Create an alarm for the 5XX metric
3. Adds a 5XX widget to the Routing Dashboard
4. Adds a 4XX widget to the Routing Dashboard

Screenshot:
Screenshot 2023-06-08 at 8 54 10 AM


bin/stacks/routing-dashboard-stack.ts
```diff
@@ -440,6 +437,27 @@ export class RoutingDashboardStack extends cdk.NestedStack {
...
+                    ['AWS/Lambda', 'Errors', 'FunctionName', ipfsPoolCacheLambdaName],
+                    ['.', 'Invocations', '.', '.'],
+                  ]
+                  : []),
+              ],
+              region: region,
+              title: 'Pool Cache Lambda Error/Invocations | 5min',
+              stat: 'Sum',
+            },
+          },
```
jsy1218, line 466: This corresponds to 429?

jsy1218: nvm, I just saw this is existing, with the positioning changes.

mikeki: I just moved that from the top to the bottom actually, I'm not sure what it means though

lib/handlers/quote/quote.ts
```diff
@@ -281,8 +281,13 @@ export class QuoteHandler extends APIGLambdaHandler<
             currencyOut.symbol
           }. Chain: ${chainId}`
         )
-
-        swapRoute = await router.route(amount, currencyOut, TradeType.EXACT_INPUT, swapParams, routingConfig)
+        try {
+          swapRoute = await router.route(amount, currencyOut, TradeType.EXACT_INPUT, swapParams, routingConfig)
+        } catch (error) {
+          metric.putMetric(`GET_QUOTE_500_CHAINID: ${chainId}`, 1, MetricLoggerUnit.Count)
```
jsy1218, line 287: For the same reason as alarming, do we want to export 5xx metrics for CHAINS_NOT_MONITORED, aka testnets?

bin/stacks/routing-api-stack.ts
```diff
@@ -363,6 +363,35 @@ export class RoutingAPIStack extends cdk.Stack {
...
+          response500: api.metric(`GET_QUOTE_500_CHAINID: ${chainId.toString()}`, {
+            period: Duration.minutes(5),
+            statistic: 'sum',
+          }),
+        },
+      })
+      const alarm = new aws_cloudwatch.Alarm(this, alarmName, {
+        alarmName,
+        metric,
+        threshold: 80,
```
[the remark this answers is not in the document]
mikeki: good call

lib/handlers/quote/quote.ts
mikeki: Let's create a ticket for that

jsy1218:
Did you have a chance to `cdk deploy` and run couple integ-test to see the SR?

lib/handlers/quote/quote.ts
```diff
@@ -42,6 +42,44 @@ export class QuoteHandler extends APIGLambdaHandler<
...
+      metric.putMetric(`GET_QUOTE_REQUESTED_CHAINID: ${chainId}`, 1, MetricLoggerUnit.Count)
+
+      switch (result.statusCode) {
+        case 200:
+        case 202:
+          metric.putMetric(`GET_QUOTE_200_CHAINID: ${chainId}`, 1, MetricLoggerUnit.Count)
+          break
+        case 400:
+        case 403:
+        case 404:
```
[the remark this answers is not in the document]
mikeki: In theory yes, but it would make the dashboard unreadable as it is right now, so I think I want to see how does it look like with just the 4XX metric first, and then consider if it's necessary to increase the level of granularity, does that sound reasonable?

lib/handlers/quote/quote.ts
```diff
@@ -42,6 +42,44 @@ export class QuoteHandler extends APIGLambdaHandler<
 > {
   public async handleRequest(
     params: HandleRequestParams<ContainerInjected, RequestInjected<IRouter<any>>, void, QuoteQueryParams>
+  ): Promise<Response<QuoteResponse> | ErrorResponse> {
+    const { chainId, metric } = params.requestInjected
+
+    let result: Response<QuoteResponse> | ErrorResponse
+
+    try {
+      result = await this.handleRequestInternal(params)
```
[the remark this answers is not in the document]
mikeki: we already track that inside the quote function, but I think it would be a good idea to bring this up here as well

lib/handlers/quote/quote.ts
mikeki: oh :facepalm: the code is already in place lol

mikeki: i think it still makes sense to have the metric here, just to see whats going in, so the default is visibility

mikeki:
> Did you have a chance to `cdk deploy` and run couple integ-test to see the SR?
Yeah I did, I've been tweaking the code based on that :p

lib/handlers/quote/quote.ts
```diff
@@ -42,6 +42,47 @@ export class QuoteHandler extends APIGLambdaHandler<
...
+    const { chainId, metric } = params.requestInjected
+    const startTime = Date.now()
+
+    let result: Response<QuoteResponse> | ErrorResponse
+
+    try {
+      result = await this.handleRequestInternal(params)
+
+      switch (result.statusCode) {
+        case 200:
```
cgkol, line 55: shall we just check for inequalities rather than specific error codes?

cgkol (commented):
Some screenshots on how this would look like on your personal account would be great too! Kind of hard to follow the code sometimes without a visual

jsy1218 (approved):
the unit test failure seems flaky. I tried it on my local and couldn't repro.

lib/handlers/quote/quote.ts
cgkol: precisely, something like that. or mayb 
if >=200 && < 300: whatever
if >=400 && < 500: lalalal

i think you get the message

mikeki: I think right now we can more easily extend if we want to have more granularity, but I dont have a strong opinion either way. I'm thinking I would merge it given the 2 approvals and then we can iterate if you feel it would be better that way

mikeki:
> I would also recommend creating one graph per p9X because it will be hard to read once everything is overlayed. Probably the same applies for average medium minimum max tbh
> 
> Thoughts?

I think you are probably right, but I was trying to find the balance between adding too many widgets and keeping grapjhs meaningful... another option would be to ignore the not monitored chains... but i kinda want to see how it would look with the actual production traffic

PR #220, opened 2023-06-08
feat: turn on rest of buckets for usdc / any token cache

lynnshaoyu:
go live for the remainder of the tick buckets for usdc/* routes. we already went live for sub 1000 USDC trades, this turns on the cache for trades > 1000 USDC.


test/unit/handlers/router-entities/route-caching/dynamo-route-caching-provider.test.ts
```diff
@@ -59,9 +59,9 @@ const TEST_WETH_USDC_POOL = new Pool(
   /* tickCurrent */ -69633
 )
 
-const TEST_USDC_UNI_POOL = new Pool(
```
mikeki, line 62: We will have to update these tests since they are not going to be maintanable in the future :p we don't need to do this now, but we will ahve to think more about it

mikeki (approved):

PR #229, opened 2023-06-12
Remove QuoteToRatio endpoint & lambda

mikeki:
## What's happening on this branch?
The QuoteToRatio endpoint (and lambdas, etc) is not being used in production, this branch is cleaning up the dead code which should simplify the service and speed up testing.

## How did I test the change?
CDK deploy and running integrationtests


cgkol (approved):
Good riddance!

mikeki:
@andysmith415 I couldn't find any reference to this endpoint there, so I think we are good

PR #236, opened 2023-06-14
Enable ConsistentRead for CachedRoutes

mikeki:
This branch enables ConsistentRead in CachedRoutes Dynamo Table.

We have the theory that this will improve cache hit rates.


mikeki:
@cgkol I think you have a good point, yes the dashboard tracks latencies for all the different pairs that have cached routes, so we will be able to notice any latency impact.

Overall I think this would be a safe-ish change and we can use it to detect wether or not there is an improvement... latency might increase, but I doubt it will be higher than infura latencies

cgkol (approved):
fair enough @mikeki

PR #261, opened 2023-07-05
[fix] [ROUTE-30] Fix RPC Provider QuickNode Metrics

jsy1218:
## What?
QuickNode RPC metrics are shown as UNKNOWN (We use QuickNode RPC for chain id 56 and 42220). We want to fix it to show QuickNode metrics.

## Why?
Because QuickNode RPC endpoint has QuikNode instead of QuickNode. We were matching against QuickNode.

## How?
Match against QuikNode.

## Testing?
N/A

## Screenshots (optional)
Screenshot 2023-07-05 at 9 15 24 AM

## Anything Else?


cgkol (approved):
Mark the PR as open!

PR #263, opened 2023-07-06
[fix] [ROUTE-34] Skip widgets for chain id with no provider

jsy1218:
## What?
Routing-API CodePipeline failed due to JSON RPC Provider URL not populated for some chain. 

## Why?
After troubleshooting, we think some net don't have the JSON RPC Provider URLs from the secrets. This was not caught in personal CDK Stack, because personal CDK Stack loads from the .env file, which adds all the chains.

## How?
We add log to ensure not all the chains are skipped. We are also not throwing error for the missing URLs to see how many URLs are missing during the codepipeline deploy.

## Testing?
Can deploy to personal CDK Stack successfully. Can see the missing chain ids in the cfn output. I intentially removed avax mainnet from my local .env. integ-test all succeeded except for avax, because of a separate change in tokenlist.

## Screenshots (optional)
Screenshot 2023-07-06 at 2 24 46 PM
Screenshot 2023-07-06 at 2 24 38 PM
Screenshot 2023-07-06 at 2 36 32 PM

## Anything Else?
N/A

### DIFF
--- a/bin/stacks/routing-dashboard-stack.ts
+++ b/bin/stacks/routing-dashboard-stack.ts
@@ -440,8 +440,10 @@ export class RoutingDashboardStack extends cdk.NestedStack {
       },
     ])
 
-    const missingProviderUrlsForChains = SUPPORTED_CHAINS.filter((chainId) => !jsonRpcProviders[`WEB3_RPC_${chainId.toString()}`])
-    const providerUrlsForChains = SUPPORTED_CHAINS.filter((chainId) => !SUPPORTED_CHAINS.includes(chainId))
+    const missingProviderUrlsForChains = SUPPORTED_CHAINS.filter(
+      (chainId) => !jsonRpcProviders[`WEB3_RPC_${chainId.toString()}`]
+    )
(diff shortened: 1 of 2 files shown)


bin/stacks/routing-dashboard-stack.ts
```diff
@@ -439,10 +440,17 @@ export class RoutingDashboardStack extends cdk.NestedStack {
       },
     ])
 
+    const missingProviderUrlsForChains = SUPPORTED_CHAINS.filter((chainId) => !jsonRpcProviders[`WEB3_RPC_${chainId.toString()}`])
+    const providerUrlsForChains = SUPPORTED_CHAINS.filter((chainId) => !SUPPORTED_CHAINS.includes(chainId))
```
mikeki, line 444: This should be `!missingProviderUrlsForChains.includes(chainId)` right?

jsy1218: yep good catch. I added the dashboard screenshot to show that the logic is correct now, otherwise no metric should show up.

mikeki (approved):

PR #270, opened 2023-07-31
Enable 1 to 2 ETH Caches as we are getting 25% of traffic for 1 ETH quotes

mikeki:
# Enable Any to ETH ExactOut Caches for 1 to 2 ETH bucket

As seen in the following screenshot, there are some bursts of requests for 1 to 2 ETH, which are not using cached routes.
In order to reduce on the cost of those requests we should enable caches.

Screenshot 2023-07-31 at 11 22 13 AM


cgkol (commented):
I think jut doing this to save money is a strategy but not the optimal one; Shall we have a more prinncipled way maybe in terms of how we roll these out moving forward?

mikeki:
> I think jut doing this to save money is a strategy but not the optimal one; Shall we have a more prinncipled way maybe in terms of how we roll these out moving forward?

Well, saving money is one reason, but overall the reason why this bucket was not enabled is because we didn't have a significant amount of traffic hitting this bucket, now that somebody is essentially quoting us with the same amount, then this bucket makes sense.

cgkol (approved):

PR #282, opened 2023-08-14
Support BaseGoerli

mikeki:


jsy1218:
It felt like when we were rushing to support `BASE`, couple places were only coded to support `BASE`, but not `BASE_GOERLI`, including this repo and in [smart-order-router](https://github.com/Uniswap/smart-order-router). 

I wonder if we first add `BASE_GOERLI` into [SUPPORTED_CHAINS](https://github.com/Uniswap/smart-order-router/blob/main/src/util/chains.ts#L4), and see if it can pass the [smart-order-router](https://github.com/Uniswap/smart-order-router) github actions integration tests? (integration tests should automatically add the BASE_GOERLI into the test case, although we might need to add JSON URL into the repo secret)

jsy1218:
I tried a bit from https://github.com/Uniswap/smart-order-router side. It appears BASE_GOERLI doesn't have the liquidity for the WETH/USDbC pools yet, so router side cannot be supported yet. Since it's testnet, we might be able to provide liquidities ourselves.

PR #283, opened 2023-08-15
Bump SOR to 3.15.4

jsy1218:
Release BASE gas bug fix https://github.com/Uniswap/smart-order-router/pull/304


jsy1218:
Close since we bump to 3.15.5. directly.

PR #285, opened 2023-08-16
Initial quote perf tuning by increasing the distribution percent from 5% to 10% for the mainnet

jsy1218:


lib/handlers/shared.ts
```diff
@@ -15,6 +15,33 @@ export const SECONDS_PER_BLOCK_BY_CHAIN_ID: { [chainId in ChainId]?: number } =
...
+          topNSecondHopForTokenAddress: new MapWithLowerCaseKey<number>([
+            ['0x5f98805a4e8be255a32880fdec7f6728c6568ba0', 2], // LUSD
+          ]),
+          topNWithEachBaseToken: 3,
+          topNWithBaseToken: 5,
+        },
+        maxSwapsPerPath: 3,
+        minSplits: 1,
+        maxSplits: 7,
+        distributionPercent: 10,
```
mikeki, line 42: I would actually just change this in the `default` section rather than making a new config for Mainnet

jsy1218: We discussed offline, and think it's a good idea to try all chains with 10% distribution.

mikeki (approved):

PR #287, opened 2023-08-16
Create quoteSpeed param and mapping

mikeki:


lib/handlers/quote/config/quote-speed-config.ts
```diff
@@ -0,0 +1,35 @@
...
+    },
+    v3PoolSelection: {
+      topN: 1,
+      topNDirectSwaps: 1,
+      topNTokenInOut: 2,
+      topNSecondHop: 0,
+      topNWithEachBaseToken: 2,
+      topNWithBaseToken: 2,
+    },
+    maxSwapsPerPath: 1,
```
willpote, line 30: Doesn't this limit us to just direct swaps?

Would put this on 2

lib/handlers/quote/config/quote-speed-config.ts
```diff
@@ -0,0 +1,35 @@
...
+      topNTokenInOut: 2,
+      topNSecondHop: 0,
+      topNWithEachBaseToken: 2,
+      topNWithBaseToken: 2,
+    },
+    v3PoolSelection: {
+      topN: 1,
+      topNDirectSwaps: 1,
+      topNTokenInOut: 2,
+      topNSecondHop: 0,
```
willpote, line 26: think we should get 1 on this, it help with ultra long tail tokens

lib/handlers/quote/config/quote-speed-config.ts
```diff
@@ -0,0 +1,35 @@
...
+      topNWithEachBaseToken: 2,
+      topNWithBaseToken: 2,
+    },
+    maxSwapsPerPath: 1,
+    maxSplits: 2,
+    distributionPercent: 10
```
willpote, line 32: If we need to I think it would be fine to raise this to 20

lib/handlers/quote/schema/quote-schema.ts
```diff
@@ -57,6 +57,7 @@ export const QuoteQueryParamsJoi = Joi.object({
   permitSigDeadline: Joi.number().optional(),
   // TODO: Remove once universal router is no longer behind a feature flag.
   enableUniversalRouter: Joi.boolean().optional().default(false),
+  quoteSpeed: Joi.string().valid('fast', 'standard', 'patient').optional().default('standard'),
```
willpote, line 60: patient is a bit of an unusual choice :D 

slow feels more standard..

lib/handlers/quote/config/quote-speed-config.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export type QuoteSpeedConfig = {
+  v2PoolSelection?: ProtocolPoolSelection;
+  v3PoolSelection?: ProtocolPoolSelection;
+  maxSwapsPerPath?: number;
+  maxSplits?: number;
+  distributionPercent?: number;
+}
+
+export const QUOTE_SPEED_MAP: { [key: string]: QuoteSpeedConfig } = {
```
willpote, line 11: https://github.com/Uniswap/routing-api/blob/98c8fbeabb8c10c93e34e2c5f2ae0d4966595805/lib/handlers/shared.ts#L16

Probably makes sense to have all the routing config in one file?

willpote (commented):
Overall looks good to me, mostly suggestions on the config for fast

lib/handlers/quote/config/quote-speed-config.ts
mikeki: I usually prefer having smaller files that are easier to read, but since we already have the pattern of having all config in one file, I guess I can follow that here.

lib/handlers/quote/schema/quote-schema.ts
mikeki: yeah, I actually think I wont encode any slower speed than 'standard', so for now only 'fast' and 'standard'

lib/handlers/quote/config/quote-speed-config.ts
jsy1218: @willpote What's your opinion on forcing direct swap for the popular pairs, like ETH/USDC? Do we need to raise the gas limit in sor?

jsy1218 (approved):

jsy1218 (approved):

PR #288, opened 2023-08-18
measure distribution change

jsy1218:
Instead of measuring how many swaps will be impacted by distribution change in [smart-order-router](https://github.com/Uniswap/smart-order-router) https://github.com/Uniswap/smart-order-router/pull/308, we should determine the impact in routing-api directly, as we plan to increase the distribution in routing-api.

### DIFF
--- a/lib/handlers/quote/quote.ts
+++ b/lib/handlers/quote/quote.ts
@@ -32,6 +32,7 @@ import { utils } from 'ethers'
 import { simulationStatusToString } from './util/simulation'
 import Logger from 'bunyan'
 import { PAIRS_TO_TRACK } from './util/pairs-to-track'
+import { measureDistributionPercentChangeImpact } from '../../util/alpha-config-measurement'
 
 export class QuoteHandler extends APIGLambdaHandler<
   ContainerInjected,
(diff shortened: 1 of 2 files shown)


lib/handlers/quote/quote.ts
```diff
@@ -549,14 +551,17 @@ export class QuoteHandler extends APIGLambdaHandler<
...
+    routeString: string,
+    swapRoute: SwapRoute
   ): void {
     const tradingPair = `${currencyIn.wrapped.symbol}/${currencyOut.wrapped.symbol}`
     const wildcardInPair = `${currencyIn.wrapped.symbol}/*`
     const wildcardOutPair = `*/${currencyOut.wrapped.symbol}`
     const tradeTypeEnumValue = tradeType == 'exactIn' ? TradeType.EXACT_INPUT : TradeType.EXACT_OUTPUT
     const pairsTracked = PAIRS_TO_TRACK.get(chainId)?.get(tradeTypeEnumValue)
 
+    measureDistributionPercentChangeImpact(5, 10, swapRoute, currencyIn, currencyOut, tradeType, chainId, amount)
```
mikeki, line 563: These numbers will get out of sync if we start changing it dynamically, so we might need to revert this branch after we collet the data we need

jsy1218: true, this is not meant to be permanent anyway.

mikeki (approved):

PR #290, opened 2023-08-18
Create a debugRoutingConfig param and a UnicornSecret

mikeki:
Tested on @jsy1218 local:

https://app.warp.dev/block/9V6k9d4rX6CWFpt18LA6SZ

### DIFF
--- a/lib/handlers/quote/quote.ts
+++ b/lib/handlers/quote/quote.ts
@@ -213,7 +213,7 @@ export class QuoteHandler extends APIGLambdaHandler<
     }
 
     let parsedDebugRoutingConfig = {}
-    if (debugRoutingConfig && unicornSecret === process.env.UNICORN_SECRET) {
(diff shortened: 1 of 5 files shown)


lib/handlers/quote/quote.ts
```diff
@@ -210,13 +212,19 @@ export class QuoteHandler extends APIGLambdaHandler<
       protocols = [Protocol.V3]
     }
 
+    let parsedDebugRoutingConfig = {}
+    if (debugRoutingConfig && unicornSecret === process.env.UNICORN_SECRET) {
```
jsy1218, line 216: I think we can be more strict - requiring both `unicornSecret` and `process.env.UNICORN_SECRET` to be populated. Because otherwise, when both are undefined, we are also parsing the `debugRoutingConfig`

jsy1218 (approved):
We also added the secret into Pipeline account

PR #293, opened 2023-08-18
Chore: measure per protocol quote latencies

jsy1218:
## What?
We are investigating deeper into the overall quote latencies. Although we have the latencies breakdown as well as overall quote latencies, we realized we don't have a breakdown of per-protocol version quote latencies breakdown. We are adding per-protocol breakdowns here.

And for a bit of context, during local perf testing, we do notice v2 quotes are actually noticeable slower than v3 quotes, if we configured local routing-api to have no cached-routes at all (https://docs.google.com/spreadsheets/d/13b-bDzOf7MQ2t_Yhu1ec2giyfe300J_rSktpUMd8HdA/edit#gid=1538815821&fvid=463061003). We do realize that `V2PoolsLoad` are impacted by the lack of caching at all (https://github.com/Uniswap/routing-api/blob/main/lib/handlers/injector-sor.ts#L265), and it always multi-call `getReserves` from v2 pools (https://github.com/Uniswap/smart-order-router/blob/main/src/providers/v2/pool-provider.ts#L112).
Screenshot 2023-08-18 at 4 54 04 PM

However, with cached routes, V2 quotes latencies are actually on-par with V3 quotes latencies. Because with cached routes, we reduces the number of `getReserves` multi-calls in v2 pool providers. This implies meanwhile adding cache on top of v2 pool provider in https://github.com/Uniswap/routing-api/blob/main/lib/handlers/injector-sor.ts#L265 is good change, it might not impact the interface/mobile quote latencies as much as we'd hope for.
Screenshot 2023-08-18 at 4 53 58 PM

## Why?
We also looked this from the latencies breakdown perspective.

At P90, `V2PoolsLoad` do look significant
Screenshot 2023-08-18 at 2 35 10 PM

But at P50, `V2PoolsLoad` no longer looks significant
Screenshot 2023-08-18 at 2 35 38 PM

This is another reason we worry why optimizing `V2PoolsLoad` might not achieve the overall quote latencies as much as we'd like.

## How?
First step, we establish overall v2/v3/mixed quoters latencies, and we can compare apple to apple that V2 quoters latencies do contribute to the latencies. If it doesn't contribute as much as v3 quoters, then we knew we weren't on the right track.

## Testing?
Tested in local routing-api, and see new stacked latencies breakdown:
Screenshot 2023-08-18 at 4 26 31 PM

## Anything Else?


mikeki (approved):

jsy1218:
Close it since we have the routes db caching optimization now

PR #300, opened 2023-08-20
Create a secondary index to query across bucket amounts

mikeki:
- Create SecondaryLocalIndex to query irrespective of buckets
- only perform search across buckets for optimistic quotes

### DIFF
--- a/bin/stacks/routing-database-stack.ts
+++ b/bin/stacks/routing-database-stack.ts
@@ -43,8 +43,9 @@ export class RoutingDatabaseStack extends cdk.NestedStack {
       timeToLiveAttribute: DynamoDBTableProps.TTLAttributeName,
     })
 
-    this.cachedRoutesDynamoDb.addLocalSecondaryIndex({
+    this.cachedRoutesDynamoDb.addGlobalSecondaryIndex({
       indexName: DynamoDBTableProps.CacheRouteDynamoDbTable.SecondaryIndexName,
+      partitionKey: { name: DynamoDBTableProps.CacheRouteDynamoDbTable.PartitionKeyName, type: AttributeType.STRING },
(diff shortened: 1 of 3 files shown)


bin/stacks/routing-database-stack.ts
```diff
@@ -41,6 +43,11 @@ export class RoutingDatabaseStack extends cdk.NestedStack {
       timeToLiveAttribute: DynamoDBTableProps.TTLAttributeName,
     })
 
+    this.cachedRoutesDynamoDb.addLocalSecondaryIndex({
```
jsy1218, line 46: Are we able to add local secondary index long after cached routes is added now?

From https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/SecondaryIndexes.html:

> Local secondary indexes are created at the same time that you create a table. You cannot add a local secondary index to an existing table, nor can you delete any local secondary indexes that currently exist.

Makes me think that we probably want to provision another test CachedRoutes table to test about the secondaryIndex change without impacting the optimistic quotes in the actual cache hit path. Since we already async invoke, we can send caching routes to second CachedRoutes table, and latencies metrics there to compare.

jsy1218 (dismissed):

bin/stacks/routing-database-stack.ts
mikeki: Good call!

I think in this case we could create a GSI instead of an LSI, I want to avoid creating another table.

jsy1218: Offline sync'ed, GSI is equivalent of creating second table (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/GSI.html#GSI.Writes), and shouldn't impact base table (CachedRoutes)

jsy1218 (approved):

PR #313, opened 2023-08-22
increase distribution percent 5 -> 10 permanently for ETH/USDC and WETH/USDC small trades

jsy1218:
## What?
We are permanently increasing the distribution percent for the most traded pairs in small trade size to speed up the `V3GetQuoteLoad` latencies, and hence E2E quote latencies on some alt-chains:
Screenshot 2023-08-20 at 9 40 54 PM

## Why?
Last time we had to revert, because we didn't know how many trade pairs will get impacted by this change. This time we have collected enough datapoints, so that we know it mostly affects the large trades.

## How?
This time we will increase the distribution percent from 5% to 10% permanently.

## Testing?
Last week we tested in prod for couple hours:
https://github.com/Uniswap/routing-api/pull/285
https://github.com/Uniswap/routing-api/pull/286

## Screenshots (optional)
We saw there was no negative impact to the quotes success rates for couple hour, so we knew this is a safe change.

## Anything Else?
N/A

### DIFF
--- a/lib/handlers/quote/quote.ts
+++ b/lib/handlers/quote/quote.ts
@@ -223,7 +223,7 @@ export class QuoteHandler extends APIGLambdaHandler<
 
     const routingConfig: AlphaRouterConfig = {
       ...DEFAULT_ROUTING_CONFIG_BY_CHAIN(chainId),
-      ...{ distributionPercent: DISTRIBUTION_PERCENT_CONFIG(chainId, currencyIn, currencyOut, type, amountRaw, log) },
+      ...{ distributionPercent: DISTRIBUTION_PERCENT_CONFIG(chainId, currencyIn, currencyOut, type, amountRaw) },
       ...(minSplits ? { minSplits } : {}),
       ...(forceCrossProtocol ? { forceCrossProtocol } : {}),
       ...(forceMixedRoutes ? { forceMixedRoutes } : {}),
--- a/lib/handlers/quote/quote.ts
+++ b/lib/handlers/quote/quote.ts
@@ -221,7 +221,7 @@ export class QuoteHandler extends APIGLambdaHandler<
     }
 
     const routingConfig: AlphaRouterConfig = {
-      ...DEFAULT_ROUTING_CONFIG_BY_CHAIN(chainId),
+      ...DEFAULT_ROUTING_CONFIG_BY_CHAIN(chainId, currencyIn, currencyOut, type, amountRaw),
       ...(minSplits ? { minSplits } : {}),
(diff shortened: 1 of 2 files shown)


lib/handlers/quote/quote.ts
```diff
@@ -221,7 +221,7 @@ export class QuoteHandler extends APIGLambdaHandler<
     }
 
     const routingConfig: AlphaRouterConfig = {
-      ...DEFAULT_ROUTING_CONFIG_BY_CHAIN(chainId),
+      ...DEFAULT_ROUTING_CONFIG_BY_CHAIN(chainId, currencyIn, currencyOut, type, amountRaw, log),
```
mikeki, line 224: I would do this override as a separate config applied on top of the default

lib/handlers/shared.ts
```diff
@@ -1,20 +1,69 @@
...
 } from '@uniswap/smart-order-router'
 import Logger from 'bunyan'
+import JSBI from 'jsbi'
 
 export const SECONDS_PER_BLOCK_BY_CHAIN_ID: { [chainId in ChainId]?: number } = {
   [ChainId.MAINNET]: 30,
 }
 
-export const DEFAULT_ROUTING_CONFIG_BY_CHAIN = (chainId: ChainId): AlphaRouterConfig => {
+export const DEFAULT_ROUTING_CONFIG_BY_CHAIN = (chainId: ChainId, currencyIn: Currency, currencyOut: Currency, type: string, amountRaw: string, log: Logger): AlphaRouterConfig => {
```
mikeki, line 18: I woudl definitely get the distribution percent logic out into a new function

jsy1218: Sure, separated into `DISTRIBUTION_PERCENT_CONFIG`

lib/handlers/quote/quote.ts
```diff
@@ -222,6 +223,7 @@ export class QuoteHandler extends APIGLambdaHandler<
 
     const routingConfig: AlphaRouterConfig = {
       ...DEFAULT_ROUTING_CONFIG_BY_CHAIN(chainId),
+      ...{ distributionPercent: DISTRIBUTION_PERCENT_CONFIG(chainId, currencyIn, currencyOut, type, amountRaw) },
```
mikeki, line 226: this will throw an error for non-eth requests.

We should either return an empty object or the object with the distributionpercent.

jsy1218:
This needs to be closed for now. We are optimizating in the fast 1st quote, which overrides to 100% distribution percent. This ties into the longer term plan of the optimistic quotes alongside the cached routes.

PR #327, opened 2023-08-25
instrument top level latencies and requests across cached routes hit vs non-cached routes hit

jsy1218:
We are instrumenting at the truly top-request level, because we suspect there's some levels of lambda cold start in the static initialization and “lazy” initialization.

Tested on personal AWS account, and see the metrics:
Screenshot 2023-08-25 at 11 25 00 AM

We are also releasing:
https://github.com/Uniswap/smart-order-router/pull/358
https://github.com/Uniswap/smart-order-router/pull/366
https://github.com/Uniswap/smart-order-router/pull/367

### DIFF
--- a/lib/handlers/handler.ts
+++ b/lib/handlers/handler.ts
@@ -108,114 +108,129 @@ export abstract class APIGLambdaHandler<CInj, RInj extends BaseRInj, ReqBody, Re
     }
   }
 
+  protected async processRequestInEntirety(
+    metric: MetricsLogger,
+    event: APIGatewayProxyEvent,
+    context: Context
+  ): Promise<[APIGatewayProxyResult, Res | null]> {
+    let log: Logger = bunyan.createLogger({
+      name: this.handlerName,
+      serializers: bunyan.stdSerializers,
+      level: bunyan.INFO,
+      requestId: context.awsRequestId,
+    })
+
+    const requestStart = Date.now()
+    log.info({ event, context }, 'Request started.')
+
+    let requestBody: ReqBody
+    let requestQueryParams: ReqQueryParams
+    try {
+      const requestValidation = await this.parseAndValidateRequest(event, log)
+
+      if (requestValidation.state == 'invalid') {
+        return [requestValidation.errorResponse, null]
+      }
+
+      requestBody = requestValidation.requestBody
+      requestQueryParams = requestValidation.requestQueryParams
+    } catch (err) {
+      log.error({ err }, 'Unexpected error validating request')
+      return [INTERNAL_ERROR(), null]
+    }
+
+    const injector = await this.injectorPromise
+
+    const containerInjected = await injector.getContainerInjected()
+
+    let requestInjected: RInj
+    try {
+      requestInjected = await injector.getRequestInjected(
+        containerInjected,
+        requestBody,
+        requestQueryParams,
+        event,
+        context,
+        log,
+        metric
+      )
+    } catch (err) {
+      log.error({ err, event }, 'Unexpected error building request injected.')
+      return [INTERNAL_ERROR(), null]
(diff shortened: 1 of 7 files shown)


lib/handlers/handler.ts
```diff
@@ -108,114 +108,129 @@ export abstract class APIGLambdaHandler<CInj, RInj extends BaseRInj, ReqBody, Re
     }
   }
 
+  protected async processRequestInEntirety(
```
mikeki, line 111: this feels dangerous, this function could be overriden by somebody else without calling super

lib/handlers/quote/quote.ts
```diff
@@ -647,4 +649,38 @@ export class QuoteHandler extends APIGLambdaHandler<
...
+    if (quoteResponseOrNull) {
+      // once the metric is here,
+      // it already got instantiated in the base handler,
+      // with the same namespace as in the quote handler
+      metric.putMetric(
+        `GET_QUOTE_LATENCY_TOP_LEVEL_${quoteResponseOrNull.hitsCache ? 'CACHED_ROUTES_HIT' : 'CACHED_ROUTES_MISS'}`,
+        Date.now() - topLevelRequestStartTime,
+        MetricLoggerUnit.Milliseconds
+      )
+      metric.putMetric(
```
mikeki, line 677: I think we only need the latency metrics, we can calculate requests from it by looking at the sample count

lib/handlers/handler.ts
```diff
@@ -108,114 +108,129 @@ export abstract class APIGLambdaHandler<CInj, RInj extends BaseRInj, ReqBody, Re
...
+      },
+      response,
+    ]
+  }
+
   private buildHandler(): APIGatewayProxyHandler {
     return metricScope(
       (metric: MetricsLogger) =>
         async (event: APIGatewayProxyEvent, context: Context): Promise<APIGatewayProxyResult> => {
-          let log: Logger = bunyan.createLogger({
```
mikeki, line 115: what if instead of doing this we create some callbacks? `beforeHandler()` and `afterHandler()` and pass some context, the metrics, logger and the response?

jsy1218: Added `afterHandler()` callback

lib/handlers/handler.ts
```diff
@@ -220,6 +224,8 @@ export abstract class APIGLambdaHandler<CInj, RInj extends BaseRInj, ReqBody, Re
     )
   }
 
+  protected abstract afterHandler(metric: MetricsLogger, response: Res, requestStart: number): void
```
mikeki, line 227: let's not make this abstract, not every handler will need to implement it. (lol we only have 1 handler here)

lib/handlers/quote/quote.ts
```diff
@@ -88,6 +89,17 @@ export class QuoteHandler extends APIGLambdaHandler<
...
+        `GET_QUOTE_LATENCY_CHAIN_${chainId}_${quoteSpeed ?? 'standard'}`,
+        Date.now() - startTime,
+        MetricLoggerUnit.Milliseconds
+      )
+      metric.putMetric(
+        `GET_QUOTE_LATENCY_CHAIN_${chainId}_${intent ?? 'quote'}`,
+        Date.now() - startTime,
+        MetricLoggerUnit.Milliseconds
+      )
```
mikeki, line 102: We have these here because of the quoteSpeed and intent params, right?

jsy1218: correct.

lib/handlers/handler.ts
jsy1218: sg, will do

mikeki (approved):

PR #330, opened 2023-08-25
Create a DDB cache for V2

mikeki:


lib/handlers/pools/pool-caching/v2/v2-dynamo-cache.ts
```diff
@@ -0,0 +1,100 @@
...
+      // Encode the jsonPair into Binary
+      const binaryPair = Buffer.from(jsonPair)
+
+      const putParams = {
+        TableName: this.tableName,
+        Item: {
+          cacheKey: key,
+          block: value.block,
+          item: binaryPair,
+          ttl: this.DEFAULT_TTL,
```
jsy1218, line 84: TTL needs to be current date + ttlMinutes right?

`Math.floor(Date.now() / 1000) + this.DEFAULT_TTL`

lib/handlers/pools/pool-caching/v2/v2-dynamo-cache.ts
```diff
@@ -0,0 +1,100 @@
...
+        base: 20,
+      },
+      httpOptions: {
+        timeout: 100,
+      },
+    })
+  }
+  async get(key: string): Promise<{ pair: Pair; block?: number } | undefined> {
+    try {
+      const queryParams = {
```
jsy1218, line 23: Feels a bit inefficient to scan the index. Think we can directly use getParam without any scanning, since we either look up when the block number is there, or not.

Like the DynamoCachingV3Pool:

```
const getParams = {
        TableName: this.tableName,
        Key: {
          poolAddress: partitionKey,
          blockNumber: sortKey,
        },
      }

      const cachedPoolBinary: Buffer | undefined = (
        await this.ddbClient
          .get(getParams)
          .promise()
          .catch((error) => {
            log.error({ error, getParams }, `[DynamoCachingV3Pool] Cached pool failed to get`)
            return undefined
          })
      )?.Item?.item
```

jsy1218 (approved):

lib/handlers/pools/pool-caching/v2/v2-dynamo-cache.ts
mikeki: Good point! thanks

lib/handlers/pools/pool-caching/v2/v2-dynamo-cache.ts
mikeki: good call

mikeki: oh I remember now... we don't have the block here

PR #335, opened 2023-08-28
Fix Cache Latency Metrics

mikeki:
#Submit a fix for the Latency metrics

I still haven't been able to confirm that the CACHE_HIT metric is working since I didnt get a cache hit locally, but I added new dashboard widgets for intent and quote speed and what I suspect is a fix for the cache hit metric

Screenshot 2023-08-28 at 9 08 00 AM


mikeki:
> Looks reasonable - seems like this line (https://github.com/Uniswap/routing-api/blob/main/lib/handlers/handler.ts#L278) removes all unknown attributes from the response body, so adding it to the response validation schema seems like it should work.
> 
> Curious - why were you unable to generate a cache hit locally?

I believe it might be due to timing, I need to send a request that would be cached after the cache gets created and before the cache expires... I could try to do it again by manually waiting a few seconds before firing a second request

PR #351, opened 2023-08-31
provision the new token properties DB for both positive and negative cache

jsy1218:
Whether we choose to use on-chain quoter or we use the token fee detector and calculate the FOT tax within smart-order-router, we will have to have a cache to avoid calculating FOT tax against non-FOT tokens. Otherwise, the additional Infura calls will be too high.

Due to the DynamoDB cost, we have decided to in-place update the token properties entry in the new table. We will persist the lastBlockSeen as part of the entry, and blocksToLive as a hardcoded value in the new `TokenProperties` Provider class. Then everytime the block seen exceeds the blocksToLive, we will in-place update the entry. Also in case there's a code bug, we hardcode each entry TTL to be 1 day (arbitrarily large) for DB growth control.

It's tested in my personal AWS account:
Screenshot 2023-09-05 at 1 13 32 PM

The dynamo caching will in addition replace the in-memory Node cache in [TokenValidatorProvider](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/token-validator-provider.ts)


bin/stacks/routing-database-stack.ts
```diff
@@ -36,6 +36,11 @@ export const DynamoDBTableProps = {
     PartitionKeyName: 'cacheKey',
     SortKeyName: 'block',
   },
+  TokenValidationResultDbTable: {
+    Name: 'TokenValidationResultDB',
+    PartitionKeyName: 'tokenAddress',
+    SortKeyName: 'chainId',
+  },
```
mikeki, line 43: I wonder if we should have:
```
paritionKey: chainIdTokenAddress,
sortKey: blockNumber
```

And we also do something like only checking every several blocknumbers?

WDYT?

jsy1218: So the way I'm thinking is that if the token is FOT/SFT, most likely they will remain that way beyond couple blocks right? So I was thinking to have a longer TTL for that. Do you see any benefit of checking every several blocknumbers, that having a longer TTL might miss?

cgkol: given tht this might be used for other properties too, I think Miguel's suggestion is a good one!

bin/stacks/routing-database-stack.ts
```diff
@@ -36,6 +36,11 @@ export const DynamoDBTableProps = {
     PartitionKeyName: 'cacheKey',
     SortKeyName: 'block',
   },
+  TokenValidationResultDbTable: {
```
cgkol, line 39: shall we maybe call it token proprties table? Given that we should be reusing this for other stuff in the future?

bin/stacks/routing-database-stack.ts
jsy1218: I adopted the feedback. Due to github issue, the PR hasn't reflected the changes yet.

jsy1218: Looks like PR reflected my commits. Ready for review again.

bin/stacks/routing-database-stack.ts
jsy1218: Renamed

cgkol (approved):

PR #354, opened 2023-09-01
Pass DDB table name from the DDB object

mikeki:


jsy1218:
From cloudwatch:

Screenshot 2023-09-01 at 10 06 14 AM

jsy1218 (approved):

PR #358, opened 2023-09-05
Reduce provisioned concurrency

mikeki:
With the recent optimizations to the routing-api, our provisioned concurrency necessary has gone down. We can reduce the minimum provisioned concurrency to save on infra cost.


cgkol (approved):
Please please monitor carefully

PR #362, opened 2023-09-12
Feature: Routing-api support for FOT quote

jsy1218:
Decent amount of changes needed in routing-api to make FOT quote work.

I tested via personal AWS account through local URA -> RA -> SOR:

https://app.warp.dev/block/dWMjnnhkKYGZRbnXf4Vy0q
Note the quote response payload from the link contains the `buyFeeBps` and `sellFeeBps` for both BLASTER and BULLET tokens.

I also added a bit of local debugging log in local sor to make sure the quote takes into account the FOT taxes:
Screenshot 2023-09-12 at 12 25 16 AM
Screenshot 2023-09-12 at 12 38 51 PM

First reserves pair quote:
BLAST => WETH
WETH amountOut = 41.968756641620631636 * 997 * 10000000000 * (1 - 0.05) / (1000 * 4896162.22883681 + 997 * 10000000000 * (1 - 0.05))
= 41.9470726768 which is close to 41.947072676794532921

Second reserves pair quote:
WETH => BULLET
BULLET amountOut = (137784644.932841276 * 997 * 41.928610626088988429 / (1000 * 4.113888521182481518 + 997 * 41.928610626088988429)) * (1 - 0.03)
= 121676691.272 which is close to 121676691.272310663

Screenshot 2023-09-12 at 12 43 05 PM
Screenshot 2023-09-12 at 12 42 57 PM

First reserves pair quote:
BLAST => WETH
WETH amountOut = 4.113888521182481518 * 0.121676691 * (1 - 0.03) * 1000 / (137784644.932841276 - 0.121676691 * (1 - 0.03)) / 997= 3.5e-9 which is close to 0.000000003756576153

Second reserves pair quote:
WETH => BULLET
BULLET amountOut = (4898323.47883681 * 0.0000000035 * 1000 / (41.950294614672082792 - 0.000000003756576153) / 997) * (1 - 0.05) = 0.0003894116 which is close to 0.00046312

Tested:
- [x] this PR unit test will continue to fail, until we publish sor changes from https://github.com/Uniswap/routing-api/pull/362
- [x] the most important aspect of this change is that `enableFeeOnTransferFeeFetching` not being passed in should make `useCachedRoutes` to be true by default:
 
```
curl 'https://ggy7t0yp57.execute-api.us-east-2.amazonaws.com/prod/quote' \
  --data-raw '{
   "tokenInChainId":1,
   "tokenIn":"0x8ef32a03784c8fd63bbf027251b9620865bd54b6",
   "tokenOutChainId":1,
   "tokenOut":"0xab98093c7232e98a47d7270ce0c1c2106f61c73b",
   "amount":"1000000000000000000",
   "type":"EXACT_INPUT",
   "configs":[
      {
         "protocols":[
            "V2",
            "V3",
            "MIXED"
         ],
         "routingType":"CLASSIC",
         "unicornSecret":"MASKED",
         "debugRoutingConfig":"{\"debugRouting\":true,\"made-up-trace-id\":\"e3d8c20e-c992-41cf-8751-23c0b49619a1\"}"
       }
   ]
}'
```
Screenshot 2023-09-13 at 8 18 13 PM


lib/handlers/marshalling/token-marshaller.ts
```diff
@@ -1,11 +1,14 @@
 import { Token } from '@uniswap/sdk-core'
+import { BigNumber } from 'ethers'
 
 export interface MarshalledToken {
   chainId: number
   address: string
   decimals: number
   symbol?: string
   name?: string
+  buyFeeBps?: BigNumber
```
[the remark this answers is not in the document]
jsy1218: we can, but curious why `string` here? `buyFeeBps` and `sellFeeBps` shouldn't have any decimals so number/BigNumber seems fine to me actually.

lib/handlers/pools/pool-caching/v2/v2-dynamo-cache.ts
```diff
@@ -18,6 +18,19 @@ export class V2DynamoCache implements ICache<{ pair: Pair; block?: number }> {
...
+    const keyValuePairs = await Promise.all(valuesPromise)
+    const records: Record<string, { pair: Pair; block?: number | undefined } | undefined> = {}
+
+    keyValuePairs.forEach(({ key, value }) => {
+      records[key] = value
+    })
+
+    return records
+  }
+
```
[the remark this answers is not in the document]
jsy1218: Sure we can. I'm thinking for the aws `batchGet` implementation, we don't implement out the handling of `unprocessedKey` to begin with, since the input token pair arrays to fetch pools caching shouldn't exceed 16MB. wdyt?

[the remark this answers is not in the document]
jsy1218: And right now sor doesn't use `batchGet` yet.

lib/handlers/marshalling/token-marshaller.ts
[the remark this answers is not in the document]
mikeki: I was thinking that for interface parsing, I dont know if its easier for them to parse a string vs checking the type

lib/handlers/pools/pool-caching/v2/v2-dynamo-cache.ts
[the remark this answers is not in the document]
mikeki: yeah... my main concern with firing multiple network calls at the same time is that it could be a bit slower, and in this case it might make sense to do batchGet since we are fetching from the same DB.

Yeah, we could add metrics to unprocessedKey just to keep track of it (and maybe set some alarm in case it ever starts showing up).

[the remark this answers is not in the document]
jsy1218: Yep, I can see the latencies concern, since we've been bite by router so bad.. Right now I think the prod has the same latencies as it could be from this PR, since it's parallelizing dynamo `get` instead of calling `batchGet` under v2 `pool-provider`. Therefore I think it makes sense to created linear ticket to follow up.

lib/handlers/marshalling/token-marshaller.ts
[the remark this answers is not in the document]
jsy1218: I'm asking internally

lib/handlers/pools/pool-caching/v2/v2-dynamo-cache.ts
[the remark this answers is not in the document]
cgkol: at the very least a TODO with a linear ticket attached

lib/handlers/quote/quote.ts
```diff
@@ -245,6 +245,12 @@ export class QuoteHandler extends APIGLambdaHandler<
       ...(enableFeeOnTransferFeeFetching ? { enableFeeOnTransferFeeFetching } : {}),
     }
 
+    if (enableFeeOnTransferFeeFetching) {
+      routingConfig.writeToCachedRoutes = true
```
cgkol, line 249: are we sure we want to enable `writeToCachedRoutes` just because we have `enableFeeOnTransferFeeFetching`? Shouldn't this be determined by the cached routes logic itself?

lib/handlers/pools/pool-caching/v2/v2-dynamo-cache.ts
```diff
@@ -18,6 +18,19 @@ export class V2DynamoCache implements ICache<{ pair: Pair; block?: number }> {
       },
     })
   }
+
+  async batchGet(keys: Set<string>): Promise<Record<string, { pair: Pair; block?: number | undefined } | undefined>> {
```
cgkol, line 22: fwiw, i dont see this function used anywhere?

lib/handlers/injector-sor.ts
```diff
@@ -190,7 +196,19 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
             sourceOfTruthPoolProvider: noCacheV3PoolProvider,
           })
 
-          const underlyingV2PoolProvider = new V2PoolProvider(chainId, multicall2Provider)
+          const tokenFeeFetcher = new OnChainTokenFeeFetcher(chainId, provider)
```
cgkol, line 199: nit: would name this `onChainTokenFeeFetcher`

lib/handlers/injector-sor.ts
```diff
@@ -351,6 +369,8 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
                 ),
                 new NodeJSCache(new NodeCache({ stdTTL: 15, useClones: false }))
               ),
+              new NodeJSCache(new NodeCache({ stdTTL: 15, useClones: false }))
```
cgkol, line 372: little confused why do we need an extra cache here?

cgkol (commented):
I assume you already have test coverage for these right?

lib/handlers/pools/pool-caching/v2/v2-dynamo-cache.ts
jsy1218: No it's not being used, hence my comment in https://github.com/Uniswap/routing-api/pull/362/files#r1323540387. @mikeki does favor using it at some point, which I agree.

lib/handlers/quote/quote.ts
jsy1218: Actually we don't need to hardcode `writeToCachedRoutes` here, because sor hardcoded the final `AlphaRouterConfig`:

```
const routingConfig: AlphaRouterConfig = _.merge(
      {
        // These settings could be changed by the partialRoutingConfig
        useCachedRoutes: true,
        writeToCachedRoutes: true,
        optimisticCachedRoutes: false,
      },
      DEFAULT_ROUTING_CONFIG_BY_CHAIN(this.chainId),
      partialRoutingConfig,
      { blockNumber }
    );
```

I think we need to hardcode `routingConfig.useCachedRoutes = false` only

lib/handlers/injector-sor.ts
jsy1218: There was a merge conflict with the hot fix stuff we just did https://github.com/Uniswap/routing-api/pull/363. We shouldn't need another in-memory cache. The only in-memory cache we added is for the TokenPropertiesProvider.

jsy1218:
> I assume you already have test coverage for these right?

I think the only missing test coverage for FOT feature overall is the missing coverage for the cached routes. The [smart-order-router](https://github.com/Uniswap/smart-order-router) integ-test won't coverbecause sor doesn't instantiate a default [routeCachingProvider](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/alpha-router.ts#L430) instance.

Also the routes caching fire and forget is issued on the routing-api side not sor side. 

So with all those considerations, we should add cached routes FOT quote testing coverage in routing-api rather than sor.

lib/handlers/quote/quote.ts
cgkol: alright shall we fix it then?

lib/handlers/pools/pool-caching/v2/v2-dynamo-cache.ts
cgkol: i would add a todo over it so people know what its about

lib/handlers/injector-sor.ts
cgkol: makes sense!

cgkol (commented):
Waiting on responses for the few left comments + integ tests and w eshould be g2g!!

lib/handlers/quote/quote.ts
```diff
@@ -245,6 +245,12 @@ export class QuoteHandler extends APIGLambdaHandler<
...
+      routingConfig.useCachedRoutes = INTENT_SPECIFIC_CONFIG[`caching`].useCachedRoutes
+      routingConfig.optimisticCachedRoutes = INTENT_SPECIFIC_CONFIG[`caching`].optimisticCachedRoutes
```
mikeki, line 251: I think here we should use `intent` instead of hardcoding `caching`?

lib/handlers/pools/pool-caching/v2/v2-dynamo-cache.ts
jsy1218: Added TODO with ROUTE-81 & ROUTE-84 so that we can triage

lib/handlers/marshalling/token-marshaller.ts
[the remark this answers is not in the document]
jsy1218: We use string

lib/handlers/pools/pool-caching/v2/v2-dynamo-cache.ts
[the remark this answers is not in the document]
jsy1218: With the additional time for dev, implemented the AWS `batchGet` here. Unfortunately, we still need a TODO with ticket, because `ICache.batchGet(keys: Set<string>)` only allows passing in simple key, meanwhile we need composite key for partition key & sort key. I created https://linear.app/uniswap/issue/ROUTE-83 to follow up.

lib/handlers/quote/quote.ts
jsy1218: Actually I think we want to use `FEE_ON_TRANSFER_SPECIFIC_CONFIG`

```
export const FEE_ON_TRANSFER_SPECIFIC_CONFIG = (enableFeeOnTransferFeeFetching?: boolean) => {
  return {
    // if enableFeeOnTransferFeeFetching is true, then we do not use cached routes for read path
    // if enableFeeOnTransferFeeFetching is false or undefined, then we use cached routes for read path
    useCachedRoutes: !enableFeeOnTransferFeeFetching,
    enableFeeOnTransferFeeFetching: enableFeeOnTransferFeeFetching,
  }
}
```

And this needs to come before `INTENT_SPECIFIC_CONFIG`, because in case someone wants to pass intent, intent should take precedence over `FEE_ON_TRANSFER_SPECIFIC_CONFIG` for `useCachedRoutes` flag

lib/handlers/quote/quote.ts
jsy1218: Yes, I actually fixed it in a better way I think:

```
export const FEE_ON_TRANSFER_SPECIFIC_CONFIG = (enableFeeOnTransferFeeFetching?: boolean) => {
  return {
    // if enableFeeOnTransferFeeFetching is true, then we do not use cached routes for read path
    // if enableFeeOnTransferFeeFetching is false or undefined, then we use cached routes for read path
    useCachedRoutes: !enableFeeOnTransferFeeFetching,
    enableFeeOnTransferFeeFetching: enableFeeOnTransferFeeFetching,
  }
}
```

Then in quote.ts:

```
const routingConfig: AlphaRouterConfig = {
      ...DEFAULT_ROUTING_CONFIG_BY_CHAIN(chainId),
      ...(minSplits ? { minSplits } : {}),
      ...(forceCrossProtocol ? { forceCrossProtocol } : {}),
      ...(forceMixedRoutes ? { forceMixedRoutes } : {}),
      protocols,
      ...(quoteSpeed ? QUOTE_SPEED_CONFIG[quoteSpeed] : {}),
      ...parsedDebugRoutingConfig,
      // enable fee-on-transfer needs to be before intent specific config,
      // in case someone wants to override useCachedRoutes with intent other than caching
      // we allow reading from the cached routes
      ...FEE_ON_TRANSFER_SPECIFIC_CONFIG(enableFeeOnTransferFeeFetching),
      ...(intent ? INTENT_SPECIFIC_CONFIG[intent] : {}),
    }
```

This way if someone wants to issue `intent` along with `enableFeeOnTransferFeeFetching`, the `useCachedRoutes` should be true, since `intent` takes precedence over `enableFeeOnTransferFeeFetching`

jsy1218:
> Waiting on responses for the few left comments + integ tests and w eshould be g2g!!

Thank you @cgkol !

I addressed all the comments, and added the integ-test to test swap BULLET -> WETH. I was hoping to be able to test BULLET -> another FOT token, but the hardhat mainnet hardfork cannot fund any other FOT token (BLASTERS, BLASTER, etc.) from the whales after impersonating their accounts during the integ-test test runtime.

Then during testing, for some reason BULLET -> WETH exact in can execute the swap on hardhat mainnet hardfork successfully, but BULLET -> WETH exact out sporadically hits the unpredictable gas limit. Since interface and mobile reconstructs the trade object and thus the transaction calldata from our routes object, I want to see if their integration would see issue with exact out FOT swap, before I call this is an issue.

lib/handlers/injector-sor.ts
cgkol: repopening since its still unresolved?

lib/handlers/quote/quote.ts
```diff
@@ -491,19 +495,41 @@ export class QuoteHandler extends APIGLambdaHandler<
               decimals: tokenIn.decimals.toString(),
               address: tokenIn.address,
               symbol: tokenIn.symbol!,
+              buyFeeBps: enableFeeOnTransferFeeFetching
```
cgkol, line 498: i would def make it a function and do if/else statements this is quite hard to read haha (i know what you are trying to do tho)

lib/handlers/shared.ts
```diff
@@ -162,6 +162,27 @@ export const INTENT_SPECIFIC_CONFIG: { [key: string]: IntentSpecificConfig } = {
   },
 }
 
+export type FeeOnTransferSpecificConfig = {
+  enableFeeOnTransferFeeFetching?: boolean
+  useCachedRoutes?: boolean
+}
+
+// TODO: ROUTE-86 - remove useCachedRoutes: !enableFeeOnTransferFeeFetching once we are ready to consume from RoutesDB
```
cgkol, line 170: perfect, so this is just for testing right? just confirming

cgkol (approved):

lib/handlers/shared.ts
jsy1218: Yep this is only used for testing. integ-test has added the [assertion](https://github.com/Uniswap/routing-api/pull/362/files#diff-e193a17cd7d38af77f287fe6bcf1c109b80c24bff4f9a4e317e4c95ea6a1c047R1054) that only in case of `enableFeeOnTransferFeeFetching` = true, we will not hit the cached routes.

lib/handlers/quote/quote.ts
jsy1218: Good point. Hard to read means it's prone to error. Also I think the copy paste here is not good. I addressed the feedback by creating private methods. Then re-run with the integ-test after re-deploying to personal aws account to ensure this works.

PR #367, opened 2023-09-13
token list provider fallback

jsy1218:


cgkol:
Not sure if there is necessary value in this wdyt? Its not that ipfs is failing its more so that we cant connect to pages in general

jsy1218:
> Not sure if there is necessary value in this wdyt? Its not that ipfs is failing its more so that we cant connect to pages in general

It's just in case if https://github.com/Uniswap/routing-api/pull/368 doesn't work, we have one more preventative measure. I think https://gateway.ipfs.io/ipns/tokens.uniswap.org 404 could be due to IPNS cannot find the IPFS hash file through the [DNSLink](https://docs.ipfs.tech/concepts/dnslink/), so the value here is to add the direct IPFS link as a fallback.

jsy1218:
> Not sure if there is necessary value in this wdyt? Its not that ipfs is failing its more so that we cant connect to pages in general

@cgkol I think we can close this one, because https://github.com/Uniswap/routing-api/pull/369 works in prod. IPFS direct fallback can be someone's starting project.

PR #375, opened 2023-09-26
feat: pass down portion

unicorn-here:
Pass down portion Options to [smart-order-router](https://github.com/Uniswap/smart-order-router)


mikeki:
Is it possible to add tests to these functions?

lib/handlers/quote/schema/quote-schema.ts
```diff
@@ -62,6 +62,16 @@ export const QuoteQueryParamsJoi = Joi.object({
...
   intent: Joi.string().valid('quote', 'swap', 'caching', 'pricing').optional().default('quote'),
   enableFeeOnTransferFeeFetching: Joi.boolean().optional().default(false),
+  portionBips: Joi.string()
+    .pattern(/^[0-9]+$/)
+    .max(5) // 0% < portionBips < 100%
+    .optional(),
+  portionAmount: Joi.string()
+    .pattern(/^[0-9]+$/)
+    .optional(),
+  portionRecipient: Joi.string()
```
zhongeric, line 72: can just use the same JOI as tokenIn address above:
`Joi.string().alphanum().max(42).optional()`

lib/handlers/shared.ts
```diff
@@ -231,3 +232,16 @@ export function parseSlippageTolerance(slippageTolerance: string): Percent {
 export function parseDeadline(deadline: string): number {
   return Math.floor(Date.now() / 1000) + parseInt(deadline)
 }
+
+export function parsePortionPercent(portionPercent: string): Percent {
```
zhongeric, line 236: if you're passing in portionBps here isn't it already denominated in terms of 10_000? I feel like you could just cast it to number.

cgkol (commented):
Definitely test the parsing part before we launch this

lib/handlers/quote/quote.ts
```diff
@@ -136,6 +137,9 @@ export class QuoteHandler extends APIGLambdaHandler<
         unicornSecret,
         intent,
         enableFeeOnTransferFeeFetching,
+        portionBips,
+        portionAmount,
```
jsy1218, line 141: Is `portionAmount` actually needed here?

lib/handlers/shared.ts
zhongeric: Yea you can still create the Percent instance but can just cast to a number instead of doing Math.floor and dividing?

lib/handlers/injector-sor.ts
```diff
@@ -204,7 +204,6 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
           )
           const tokenPropertiesProvider = new TokenPropertiesProvider(
             chainId,
-            tokenValidatorProvider,
```
cgkol, line 207: is this relic from a previous PR?

lib/handlers/shared.ts
```diff
@@ -164,21 +165,12 @@ export const INTENT_SPECIFIC_CONFIG: { [key: string]: IntentSpecificConfig } = {
...
-// because RoutesDB has 24hrs TTL, and routing-api no longer filters unexpired routers.
-// So during interface & mobile e2e testing, it won't work if the fot quote hits the cached routes read path.
-// We allow writing into RoutesDB but not reading from it, if enableFeeOnTransferFeeFetching is true.
 export const FEE_ON_TRANSFER_SPECIFIC_CONFIG = (
   enableFeeOnTransferFeeFetching?: boolean
 ): FeeOnTransferSpecificConfig => {
   return {
-    // if enableFeeOnTransferFeeFetching is true, then we do not use cached routes for read path
-    // if enableFeeOnTransferFeeFetching is false or undefined, then we use cached routes for read path
-    useCachedRoutes: !enableFeeOnTransferFeeFetching,
```
cgkol, line 181: this should be from an old PR as well right?

cgkol (approved):
✅

jsy1218 (approved):

PR #389, opened 2023-10-11
fix: routing-api doesn't own the portion bps/amount default value

ghost:
routing-api should not own the portion bps/amount default value. URA should inside https://github.com/Uniswap/unified-routing-api/blob/main/lib/providers/quoters/RoutingApiQuoter.ts#L88-L102


cgkol (approved):
So the adding 0 if nothing is there, is added on URA side right?

PR #404, opened 2023-11-03
Switch to StaticJsonRpcProvider

mikeki:
From this issue in ethers: https://github.com/ethers-io/ethers.js/issues/901

We can switch to using the StaticJsonRpcProvider which avoids an unnecessary chainId call to infura, saving about 30% of requests to infura


zzmp (commented):
It's not obvious to me that this provider will only ever be used for the chain for which it is initialized. Is that true? You can also pass in add'l info in initializing to prevent it calling to the chain for chainInfo.

zzmp (approved):

zhongeric (approved):

PR #406, opened 2023-11-03
Cache eth_calls in our InstrumentedEVMProvider

mikeki:
Bsed of the [interface's](https://github.com/Uniswap/interface/blob/main/src/rpc/StaticJsonRpcProvider.ts#L31-L52) caching mechanism, we are also adding the caching capabilities to our rpv provider


cgkol (approved):
🔥

zzmp (approved):
lgtm! do you need a unit tst?

PR #416, opened 2023-11-09
Only send Caching requests every X blocks

mikeki:
This change will change the behavior of how often we sample buckets in RoutesDB.

We currently have a RoutesDB that stores routes seen for the past 24 hours for every specific pair. This database is not really growing much, since most routes will tend to stay liquid within the same liquidity pools that are currently being traded.

We care about updating the database whenever theres a major change in liquidity, but that doesn't happen often, so we are setting some threshold for how often we sample each *bucket* of a given pair. (A bucket is a range of from amount to amount * phi), we are changing from sampling on every block to sampling less often to reduce the amount of sync request we send to the routing api.

We will likely have to play with the numbers based on the chain velocity (how often a block is generated).

### DIFF
--- a/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
+++ b/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
@@ -83,10 +83,10 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
   private readonly ROUTES_DB_BUCKET_RATIO: Fraction = new Fraction(514229, 317811)
   private readonly ROUTES_TO_TAKE_FROM_ROUTES_DB = 8
   private readonly BLOCKS_DIFF_BETWEEN_CACHING_QUOTES: Map<ChainId, number> = new Map([
-    [ChainId.MAINNET, 10]
+    [ChainId.MAINNET, 3]
   ])
 
-  private readonly DEFAULT_BLOCKS_DIFF_CACHING = 100;
+  private readonly DEFAULT_BLOCKS_DIFF_CACHING = 15;
 
   constructor({ routesTableName, routesCachingRequestFlagTableName, cachingQuoteLambdaName }: ConstructorParams) {
     super()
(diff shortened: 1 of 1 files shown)


lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
```diff
@@ -300,8 +303,13 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
       const shouldSendCachingRequest =
         result.Items &&
         (result.Items.length == 0 || // no caching request has been sent recently
-          // or there is not a single sample for the current block
-          result.Items.every((record) => (record.blockNumber ?? 0) < currentBlockNumber))
+          // or every sampled record is older than maximum blocks diff allowed for the chain
+          result.Items.every((record) => {
+            const blocksDiff = currentBlockNumber - (record.blockNumber ?? 0)
+            const maximumBlocksDiff =
+              this.BLOCKS_DIFF_BETWEEN_CACHING_QUOTES.get(partitionKey.chainId) || this.DEFAULT_BLOCKS_DIFF_CACHING
```
cgkol, line 310: for now `BLOCKS_DIFF_BETWEEN_CACHING_QUOTES` is always  null so we defualt to `DEFAULT_BLOCKS_DIFF_CACHING` right?

cgkol (approved):

lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
mikeki: that one actually contains Mainnet: 3

mikeki: https://github.com/Uniswap/routing-api/pull/416/files#diff-ba752e09db66ef293edd546c80fadaf47bb160d3542dca3c31cbbb3628bd0cdcR85

cgkol: i cant read lol

PR #424, opened 2023-11-16
Chore: add tenderly simulation http status alarm

jsy1218:
As a follow-up action item from https://www.notion.so/uniswaplabs/New-tokens-swap-outages-on-Wallet-due-to-Tenderly-enforcing-HTTPS-aad4719087f0407d8a9b21f0137fd070?pvs=4#86b9072133b949a6b9e636dfcca1e76e, we will need to be able to detect and alert, when the tenderly requests are no longer returning success (http status 200).

It makes sense to alert against universal router simulation, but not swap router 02, due to swap router 02 being intended to get deprecated. Also universal router simulation is the main simulation traffic volume.

1. If tenderly no longer returns HTTP 200, then there will be insufficient datapoints, turning into alarm, tested in local AWS account: Screenshot 2023-11-16 at 3 34 30 PM
2. If there's enough datapoints, every 5 minutes the evaluation should result in more than 99% in prod:
Screenshot 2023-11-16 at 3 45 02 PM


mikeki (approved):
linter is failing though

PR #435, opened 2023-12-04
Make `source` param optional in quote API

rollingtumbling:
Since not all the clients have added the `source` param in their request, we have to make it optional to unblock deployment.


dannydaniil (approved):
lgtm

PR #437, opened 2023-12-14
Pass down gas token to routingConfig for relay quote support

zhongeric:
Pass down optional `gasToken` parameter (token address) to smart order router and returns the additional response fields if present. 

Fully backward compatible as classic quotes without gasToken specified just don't return the fields.


jsy1218 (approved):
We should probably add an test in integ-test, i.e. in quote.test.ts, for passing in the gas token.

jsy1218 (approved):

PR #449, opened 2024-01-03
RPC Provider Gateway

rollingtumbling:
## Feature
Please see [Design doc](https://www.notion.so/uniswaplabs/RPC-Proxy-for-Providers-5dbb8faf379948c0bfa03a28a94732b2?pvs=4). Here are more implementation notes not covered in the design doc:

### Two layer of providers
For implementation,  we introduced a `UniJsonRpcProvider` and `SingleJsonRpcProvider`
- `UniJsonRpcProvider`: Serves as the gateway to select the right provider
- `SingleJsonRpcProvider`: Serve the actual request. Also maintain the health information of this provider

Most etheres APIs are wrapped to provide the same user experience.

### Singleton / Life cycle
We recommend user to get the provider instance by `GlobalRpcProviders.getGlobalUniRpcProviders()`. This allows all user code to share the same "health situation" for these providers, so when one provider failed, all user code will know about it immediately. 

However, if a piece of user code want to use different provider selection strategy, he can spawn a new `UniJsonRpcProvider`, but still recommended to take `GlobalRpcProviders.getGlobalSingleRpcProviders()` to use the shared `SingleJsonRpcProvider`s to share the provider health info with other user code.
 
### Session support
`UniJsonRpcProvider` supports creating new session ID. Using the same session ID ensure there's no provider switch (otherwise provider may switch on any given call).

### Unit test
Added unit test to cover all introduced logic.
Since there are a lot of hand-copied code due to API wrapping, we only use `getBlockNumber()` as an example API to test in the unit tests. The current test coverage is
Screenshot 2024-01-08 at 2 29 18 AM

But if we exclude the verbose API wrapper code, the UT coverage for both `uniJsonRpcProvider.ts` and `singleJsonRpcProvider.ts` are over 95%.

### Manual integration test
For Ethereum mainnet, we replace the current provider in use `InstrumentedEVMProvider` and use `UniJsonRpcProvider` instead. Then we curl the `quote` request like this

```
curl --location 'https://k0g4phvg10.execute-api.us-east-2.amazonaws.com/prod/quote?tokenInAddress=0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2&tokenInChainId=1&tokenOutAddress=0x1f9840a85d5af5bf1d1762f925bdaddc4201f984&tokenOutChainId=1&amount=100&type=exactIn'
```
And the result looks normal
```
{
    "blockNumber": "18959846",
    "amount": "100",
    "amountDecimals": "0.0000000000000001",
    "quote": "37588",
    "quoteDecimals": "0.000000000000037588",
    "quoteGasAdjusted": "-1497465077425683691",
    "quoteGasAdjustedDecimals": "-1.497465077425683691",
    "gasUseEstimateQuote": "1497465077425721279",
    "gasUseEstimateQuoteDecimals": "1.497465077425721279",
    "gasUseEstimate": "128000",
    "gasUseEstimateUSD": "8.629910639342568964",
    "simulationStatus": "UNATTEMPTED",
    "simulationError": false,
    "gasPriceWei": "30812169074",
    "route": [
        [
            {
                "type": "v3-pool",
                "address": "0x1d42064Fc4Beb5F8aAF85F4617AE8b3b5B8Bd801",
                "tokenIn": {
                    "chainId": 1,
                    "decimals": "18",
                    "address": "0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2",
                    "symbol": "WETH"
                },
                "tokenOut": {
                    "chainId": 1,
                    "decimals": "18",
                    "address": "0x1f9840a85d5aF5bf1D1762F925BDADdC4201F984",
                    "symbol": "UNI"
                },
                "fee": "3000",
                "liquidity": "327862648197731747445127",
                "sqrtRatioX96": "4065999802749605505505689923",
                "tickCurrent": "-59397",
                "amountIn": "100",
                "amountOut": "37588"
            }
        ]
    ],
    "routeString": "[V3] 100.00% = WETH -- 0.3% [0x1d42064Fc4Beb5F8aAF85F4617AE8b3b5B8Bd801] --> UNI",
    "quoteId": "27a47",
    "hitsCachedRoutes": false
}
```
### E2E integration test
We run `npm run test:e2e`, all tests pass
Screenshot 2024-01-08 at 1 37 54 AM

### Percentage-based routing test
We try to balance traffic between Infura and Alchemy at 1:1 ratio and retry above E2E integration test. 
It works as expected.

### Fallback test
We set up 1 failing providers + 1 healthy providers (in that preference order). First one will fallback after a while, then use the backup to serve. So the results should be failed requests first(served by 1st provider), then followed by succeeded requests(served by 2nd provider).
It works as expected.

### Known issues
Currently we store health states in lambda's local memory. Since AWS can spawn new instance or reuse existing lambda instance arbitrarily, we don't have a control for the life cycle of the provider health states. To address this issue, we need to use DynamoDB table instead. This is important and needs to be done in the next PR.

#### Policy to report error
Currently when any of the etheres API throws an error, we will record this as a provider error and deduct the health score. However I'm not quite sure this is fair to the provider. For example, I've seen this error during manual integration test
Screenshot 2024-01-08 at 1 57 37 AM

This will for sure throws an error, but is this a provider's issue? We use http 4xx code to indicate error from user provided data and 5xx for server internal error, but I'm not sure we can differentiate that from etheres API's response. Please help!


lib/handlers/injector-sor.ts
```diff
@@ -53,6 +52,9 @@ import { deriveProviderName } from './evm/provider/ProviderName'
 import { V2DynamoCache } from './pools/pool-caching/v2/v2-dynamo-cache'
 import { OnChainTokenFeeFetcher } from '@uniswap/smart-order-router/build/main/providers/token-fee-fetcher'
 import { PortionProvider } from '@uniswap/smart-order-router/build/main/providers/portion-provider'
+import GlobalRpcProviders from '../rpc/globalRpcProviders'
+import { StaticJsonRpcProvider } from '@ethersproject/providers'
+// import UniJsonRpcProvider from '../rpc/uniJsonRpcProvider'
```
cgkol, line 57: delete comment?

lib/rpc/config.ts
```diff
@@ -0,0 +1,19 @@
...
+  HIGH_LATENCY_PENALTY: number
+  HEALTH_SCORE_FALLBACK_THRESHOLD: number
+  HEALTH_SCORE_RECOVER_THRESHOLD: number
+  MAX_LATENCY_ALLOWED_IN_MS: number
+  RECOVER_SCORE_PER_MS: number
+  RECOVER_EVALUATION_WAIT_PERIOD_IN_MS: number
+}
+
+export const DEFAULT_CONFIG: Config = {
+  ERROR_PENALTY: -50,
```
cgkol, line 12: might be worth commenting what these means; in other words, if we accumulate up to 3990 penalties in 5seconds then we fallback; I assume thats the case, and you can also explain the recovery part too

lib/rpc/singleJsonRpcProvider.ts
```diff
@@ -0,0 +1,460 @@
...
+        this.checkLastCallPerformance('getTransaction')
+      })
+  }
+
+  override getTransactionCount(
+    addressOrName: string | Promise<string>,
+    blockTag?: BlockTag | Promise<BlockTag>
+  ): Promise<number> {
+    const startTime = Date.now()
+    this.recordPerfBeforeCall(startTime)
```
cgkol, line 299: i see there is a lot of duplication around 

```
    const startTime = Date.now()
    this.recordPerfBeforeCall(startTime)
```

Would it make more sense to make all these super calls encapsulated somehow so that we dont do this everytime? I think this might make this a lot more readable! We have certain examples on data-api-graphql

cgkol: practically, the only variable is the function call (`getTransactionCount` or `getTransactionReceipt` etc) so we can do something cool here

lib/rpc/singleJsonRpcProvider.ts
```diff
@@ -0,0 +1,460 @@
...
+      this.isRecovering = true
+      debug(`${this.url} drops to unhealthy`)
+    } else if (this.isRecovering && this.healthScore > this.config.HEALTH_SCORE_RECOVER_THRESHOLD) {
+      this.isRecovering = false
+      debug(`${this.url} resumes to healthy`)
+    }
+    // No reward for normal operation.
+  }
+
+  private recordPerfBeforeCall(startTimeInMs: number) {
```
cgkol, line 104: here you need to update the startTimeInMs too right?

cgkol: i see you pass it as an argument on the afterCall function, but probably a little more nebulous if you do it this way

lib/rpc/singleJsonRpcProvider.ts
```diff
@@ -0,0 +1,460 @@
...
+      metric.putMetric(`${this.metricPrefix}_${method}_FAILED`, 1, MetricLoggerUnit.Count)
+      this.recordError()
+    } else if (this.perf.lastCallLatencyInMs > this.config.MAX_LATENCY_ALLOWED_IN_MS) {
+      metric.putMetric(`${this.metricPrefix}_${method}_SUCCESS_HIGH_LATENCY`, 1, MetricLoggerUnit.Count)
+      this.recordHighLatency()
+    } else {
+      metric.putMetric(`${this.metricPrefix}_${method}_SUCCESS`, 1, MetricLoggerUnit.Count)
+      // For a success call, we will increase health score.
+      if (this.perf.timeWaitedBeforeLastCallInMs > 0) {
+        this.recordProviderRecovery(this.perf.timeWaitedBeforeLastCallInMs)
```
cgkol, line 91: so we increase proportionally to the time between the last call? whats the intuition behind this?

cgkol: The longer we wait the more likely for it to have recovered, given that the last call was succesful?

lib/rpc/singleJsonRpcProvider.ts
```diff
@@ -0,0 +1,460 @@
...
+    super(url, { chainId, name: CHAIN_IDS_TO_NAMES[chainId] })
+    this.url = url
+    this.healthScore = 0
+    this.isRecovering = false
+    this.perf = new PerfStat()
+    this.config = config
+    this.metricPrefix = `RPC_${this.network.chainId}_${this.url}`
+  }
+
+  isHealthy(): boolean {
```
cgkol, line 46: we dont use isHealthy anywhere?

lib/rpc/chains.ts
```diff
@@ -0,0 +1,21 @@
+import { ChainId } from '@uniswap/sdk-core'
+
+export const CHAIN_IDS_TO_NAMES = {
```
cgkol, line 3: btw dont we already have such a conversion? otherwise how do we pass in the chain currently?

bin/app.ts
```diff
@@ -281,6 +281,9 @@ const jsonRpcProviders = {
   WEB3_RPC_56: process.env.JSON_RPC_PROVIDER_56!,
   WEB3_RPC_43114: process.env.JSON_RPC_PROVIDER_43114!,
   WEB3_RPC_8453: process.env.JSON_RPC_PROVIDER_8453!,
+  UNI_RPC_PROVIDER_INFURA_KEY: process.env.UNI_RPC_PROVIDER_INFURA_KEY!,
```
jsy1218, line 284: I think we should name it more generically, because we might dynamically switch between providers at runtime.

lib/rpc/chains.ts
jsy1218: Yea, routing-api imports `ID_TO_NETWORK_NAME` from sor

lib/rpc/globalRpcProviders.ts
```diff
@@ -0,0 +1,62 @@
...
+  private static initGlobalSingleRpcProviders() {
+    const INFURA_KEY = process.env.UNI_RPC_PROVIDER_INFURA_KEY
+    if (INFURA_KEY === undefined) {
+      throw new Error(`UNI_RPC_PROVIDER_INFURA_KEY must be a defined environment variable`)
+    }
+    const QUICKNODE_MAINNET_RPC_URL = process.env.UNI_RPC_PROVIDER_QUICKNODE_MAINNET_RPC_URL
+    if (QUICKNODE_MAINNET_RPC_URL === 'undefined') {
+      throw new Error(`REACT_APP_QUICKNODE_MAINNET_RPC_URL must be a defined environment variable`)
+    }
+    this.SINGLE_RPC_PROVIDERS = {
```
jsy1218, line 27: This implementation will make the RPC provider gateway to roll out to ETH Mainnet completely. I think we want to de-risk by deploying to a chain with lesser usages/revenues, meanwhile we know we are using the Infura RPC for that chain.

By looking at the Infura RPCs, and the chains we collect revenue on, Avalanche 43114 might be a good choice.

lib/rpc/globalRpcProviders.ts
```diff
@@ -0,0 +1,62 @@
+import { ChainId } from '@uniswap/sdk-core'
+import SingleJsonRpcProvider from './singleJsonRpcProvider'
+import UniJsonRpcProvider from './uniJsonRpcProvider'
+
+export default class GlobalRpcProviders {
+  private static readonly PROVIDER_RPC_URL_RANKING: Partial<Record<ChainId, number[] | undefined>> = {
```
jsy1218, line 6: It feels like we can just use `Map` construct instead of `Partial<Record>`. There are places in routing-api where we use Map, like https://github.com/Uniswap/routing-api/blob/main/lib/handlers/quote/util/pairs-to-track.ts

lib/rpc/globalRpcProviders.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  private static UNI_RPC_PROVIDERS: Partial<Record<ChainId, UniJsonRpcProvider>> | null = null
+
+  private static initGlobalSingleRpcProviders() {
+    const INFURA_KEY = process.env.UNI_RPC_PROVIDER_INFURA_KEY
+    if (INFURA_KEY === undefined) {
+      throw new Error(`UNI_RPC_PROVIDER_INFURA_KEY must be a defined environment variable`)
+    }
+    const QUICKNODE_MAINNET_RPC_URL = process.env.UNI_RPC_PROVIDER_QUICKNODE_MAINNET_RPC_URL
+    if (QUICKNODE_MAINNET_RPC_URL === 'undefined') {
```
jsy1218, line 24: I think you need to remove the quote for checking definedness. Otherwise you are checking `UNI_RPC_PROVIDER_QUICKNODE_MAINNET_RPC_URL` is not string literal `undefined`.

lib/rpc/globalRpcProviders.ts
```diff
@@ -0,0 +1,62 @@
...
+        GlobalRpcProviders.PROVIDER_RPC_URL_WEIGHTS[ChainId.MAINNET]
+      ),
+    }
+  }
+
+  static getGlobalSingleRpcProviders(): Partial<Record<ChainId, SingleJsonRpcProvider[]>> {
+    if (this.SINGLE_RPC_PROVIDERS === null) {
+      this.initGlobalSingleRpcProviders()
+    }
+    return this.SINGLE_RPC_PROVIDERS!
```
jsy1218, line 53: nit: the code should be able to get simplified using null coalesce

```
return this.SINGLE_RPC_PROVIDERS ?? this.initGlobalSingleRpcProviders();
```

bin/app.ts
```diff
@@ -281,6 +281,9 @@ const jsonRpcProviders = {
...
+  UNI_RPC_PROVIDER_QUICKNODE_MAINNET_RPC_URL: process.env.UNI_RPC_PROVIDER_QUICKNODE_MAINNET_RPC_URL!,
+  DEBUG: process.env.DEBUG!,
```
jsy1218, line 286: How is this DEBUG being used?

package.json
```diff
@@ -12,7 +12,7 @@
     "build": "run-p compile-external-types compile-v3-contract-types && tsc",
     "clean": "rm -rf dist cdk.out",
     "watch": "tsc -w",
-    "test:unit": "jest test/jest/unit/",
+    "test:unit": "jest test/jest/unit/ && ts-mocha -p tsconfig.cdk.json lib/rpc/**/*.test.ts",
```
jsy1218, line 15: yea, I think you need to move the tests to `test/jest/unit/`, and then you don't need to modify the `test:unit` script

package.json
```diff
@@ -30,6 +30,7 @@
     "@types/aws-lambda": "^8.10.77",
     "@types/bunyan": "^1.8.6",
     "@types/chai-as-promised": "^7.1.4",
+    "@types/debug": "^4.1.12",
```
jsy1218, line 33: Is this related to the DEBUG env variable added above? Do we actually need to add this dependency?

package-lock.json
```diff
@@ -21236,6 +21504,264 @@
...
+        "node": ">=8"
+      }
+    },
+    "node_modules/nyc/node_modules/y18n": {
+      "version": "4.0.3",
+      "resolved": "https://registry.npmjs.org/y18n/-/y18n-4.0.3.tgz",
+      "integrity": "sha512-JKhqTOwSrqNA1NY5lSztJ1GrBiUodLMmIZuLiDaMRJ+itFd+ABVE8XBjOvIWL+rSqNDC74LCSFmlb/U4UZ4hJQ==",
+      "dev": true
+    },
+    "node_modules/nyc/node_modules/yargs": {
```
jsy1218, line 21730: wow that's a lot of dep changes in lock file. Whenever I see a lot of dep changes in lock file, I'd be cautious on whether all those dep changes are necessary.

lib/rpc/uniJsonRpcProvider.ts
```diff
@@ -0,0 +1,496 @@
...
+  Log,
+  TransactionReceipt,
+  TransactionResponse,
+} from '@ethersproject/abstract-provider'
+import { LRUCache } from 'lru-cache'
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber'
+import { Network } from '@ethersproject/networks'
+import { Deferrable } from '@ethersproject/properties'
+
+const debug = Debug('UniJsonRpcProvider')
```
jsy1218, line 19: We have log.debug using bunyan library.

Then when you want to enable debug on your local, you can modify the log level to debug at https://github.com/Uniswap/routing-api/blob/main/lib/handlers/quote/injector.ts#L38.

lib/rpc/singleJsonRpcProvider.ts
```diff
@@ -0,0 +1,460 @@
...
+      .catch((error) => {
+        callSucceed = false
+        throw error
+      })
+      .finally(() => {
+        const endTime = Date.now()
+        this.recordPerfAfterCall(startTime, endTime, callSucceed)
+        this.checkLastCallPerformance('send')
+      })
+  }
```
jsy1218, line 459: I think we don't need to override all the JSON RPC methods, because many of them are not invoked in prod today.

Originally we decided to override every JSON RPC method in [InstrumentedEVMProvider](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/evm/provider/InstrumentedEVMProvider.ts), so that we can see which methods are being invoked in routing-api runtime, although we already knew some methods are not being invoked like `getLogs`, because we don't have a use case to inspect ETH logs.

You can use the [AWS metrics](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#metricsV2?graph=~(metrics~(~(~'Uniswap~'RPC_INFURA_1_RESOLVENAME_REQUESTED~'Service~'RoutingAPI)~(~'.~'RPC_INFURA_1_GETNETWORK_REQUESTED~'.~'.)~(~'.~'RPC_INFURA_1_ESTIMATEGAS_SUCCESS~'.~'.)~(~'.~'RPC_INFURA_1_CALL_LATENCY~'.~'.)~(~'.~'RPC_INFURA_1_ESTIMATEGAS_REQUESTED~'.~'.)~(~'.~'RPC_INFURA_1_ESTIMATEGAS_FAILURE~'.~'.)~(~'.~'RPC_INFURA_1_CALL_FAILURE~'.~'.)~(~'.~'RPC_INFURA_1_GETBLOCKNUMBER_REQUESTED~'.~'.)~(~'.~'RPC_INFURA_1_CALL_REQUESTED~'.~'.)~(~'.~'RPC_INFURA_1_CALL_SUCCESS~'.~'.)~(~'.~'RPC_INFURA_1_GETNETWORK_SUCCESS~'.~'.)~(~'.~'RPC_INFURA_1_GETBLOCKNUMBER_FAILURE~'.~'.)~(~'.~'RPC_INFURA_1_GETBLOCKNUMBER_SUCCESS~'.~'.)~(~'.~'RPC_INFURA_1_RESOLVENAME_SUCCESS~'.~'.)~(~'.~'RPC_INFURA_43114_GETNETWORK_REQUESTED~'.~'.)~(~'.~'RPC_INFURA_43114_GETNETWORK_SUCCESS~'.~'.)~(~'.~'RPC_INFURA_43114_GETBLOCKNUMBER_FAILURE~'.~'.)~(~'.~'RPC_INFURA_43114_CALL_LATENCY~'.~'.)~(~'.~'RPC_INFURA_43114_RESOLVENAME_SUCCESS~'.~'.)~(~'.~'RPC_INFURA_43114_CALL_FAILURE~'.~'.)~(~'.~'RPC_INFURA_43114_GETBLOCKNUMBER_REQUESTED~'.~'.)~(~'.~'RPC_INFURA_43114_GETGASPRICE_FAILURE~'.~'.)~(~'.~'RPC_INFURA_43114_GETGASPRICE_SUCCESS~'.~'.)~(~'.~'RPC_INFURA_43114_GETBLOCKNUMBER_SUCCESS~'.~'.)~(~'.~'RPC_INFURA_43114_RESOLVENAME_REQUESTED~'.~'.)~(~'.~'RPC_INFURA_43114_GETGASPRICE_REQUESTED~'.~'.)~(~'.~'RPC_INFURA_43114_CALL_SUCCESS~'.~'.)~(~'.~'RPC_INFURA_43114_CALL_REQUESTED~'.~'.))~view~'timeSeries~stacked~false~region~'us-east-2~start~'-PT1H~end~'P0D~stat~'SampleCount~period~300)&query=~'*7bUniswap*2cService*7d*20RPC_INFURA_43114) to see which methods are invoked today. For example, I want to check mainnet and avalanche, so I put prefix `RPC_INFURA_1` and `RPC_INFURA_43114`, and see:

Screenshot 2024-01-10 at 11 11 24 AM

In fact, the [Routing Dashboard](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#dashboards/dashboard/RoutingDashboard) only show the JSON RPC methods that are invoked at the bottom of the page, because majority of the JSON RPC methods are not invoked.

lib/rpc/singleJsonRpcProvider.ts
```diff
@@ -0,0 +1,460 @@
...
+  private readonly metricPrefix: string
+
+  constructor(chainId: ChainId, url: string, config: Config = DEFAULT_CONFIG) {
+    super(url, { chainId, name: CHAIN_IDS_TO_NAMES[chainId] })
+    this.url = url
+    this.healthScore = 0
+    this.isRecovering = false
+    this.perf = new PerfStat()
+    this.config = config
+    this.metricPrefix = `RPC_${this.network.chainId}_${this.url}`
```
jsy1218, line 43: Seeing this metrics prefix, I was worried about the API Key part of the RPC URL will get exported to the metrics. Checking with https://github.com/Uniswap/routing-api/pull/449/files#diff-e594d19771a20ac88f1ab3d2795c11801460d30a4d14aad748a4ba55987865f4R29-R30, that indeed seems like the case. 

I think we still want to call [deriveProviderName](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/evm/provider/ProviderName.ts#L8) to have a more human-friendly as well as shortened metrics prefix.

lib/rpc/singleJsonRpcProvider.ts
```diff
@@ -0,0 +1,460 @@
...
+        timeInMs * this.config.RECOVER_SCORE_PER_MS
+      }, score => ${this.healthScore}`
+    )
+  }
+
+  private checkLastCallPerformance(method: string) {
+    if (!this.perf.lastCallSucceed) {
+      metric.putMetric(`${this.metricPrefix}_${method}_FAILED`, 1, MetricLoggerUnit.Count)
+      this.recordError()
+    } else if (this.perf.lastCallLatencyInMs > this.config.MAX_LATENCY_ALLOWED_IN_MS) {
```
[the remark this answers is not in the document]
jsy1218: I think we will need per-call granularity. In addition, we will also need per-chain granularity. Practically, when we see an infura outage, it could be a partial outage in the form of some RPC calls failure and/or some chains outage.

jsy1218 (commented):
Which RPC methods to override, how to override RPC methods for health gauge, as well as which chain(s) to roll out require more thinking.

In general, I'd think from the following perspectives:

- What are the top JSON RPC calls to Infura?

Below is the screenshot back in Nov:
Screenshot 2023-11-13 at 5 08 14 PM

Worth noting couple points:

1. eth_call although it's dominant, it might not be a direct invocation to [InstrumentedEVMClient.call()](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/evm/provider/InstrumentedEVMProvider.ts#L35). It could be a multicall within [UniswapMulticallProvider](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/multicall-uniswap-provider.ts#L33), which ethers v5 library internally will invoke  [InstrumentedEVMClient.call()](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/evm/provider/InstrumentedEVMProvider.ts#L35) eventually. This means if we gauge eth_call at [InstrumentedEVMClient.call()](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/evm/provider/InstrumentedEVMProvider.ts#L35), meanwhile a single multicall failed due to the same reason, which spins multiple eth_calls, we might inflate the provider's failure rate here.
2. eth_blockNumber is usually only invoked at the beginning of SOR [alpha-router](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/alpha-router.ts#L2384), because we usually pass the block number to the block tag for the subsequent RPC calls, e.g. [multicall block tag](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/multicall-uniswap-provider.ts#L96). And since block number is available in each block header (meaning this data piece is from a lightweight component), `getBlockNumber()` might actually be a good candidate to start trying out the health gauge with.
3. eth_chainID is not explicitly invoked anywhere in routing-api or SOR. Also since RPC URL is per chain, eth_chainID is not an ideal candidate.
4. eth_feeHistory and eth_gasPrice might also be an ideal candidate, although it's only for mainnet. eth_feeHistory is only applicable to [eip-1559-gas-provider](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/eip-1559-gas-price-provider.ts), on mainnet. As for gas price, L2 might have its own gas price oracles, that doesn't invoke `getGasPrice()` here, e.g. Arbitrum [GasInfo](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/v3/gas-data-provider.ts#L115), which calls out to Arbitrum's gas oracles [contract](https://docs.arbitrum.io/for-devs/dev-tools-and-resources/precompiles#arbgasinfo).

- In the last couple quote endpoint incidents due to Infura (not due to the underlying network issue), can we derive a pattern?
This will require us to look into [#x-infura](https://uniswapteam.slack.com/archives/C023A7JDTJP) as well as [#pagerduty-routing](https://uniswapteam.slack.com/archives/C059TGEC57W). Notably there are 3 incidents that are purely due to Infura components, but not due to the underlying network issue:

1. Nov 30th [one](https://uniswapteam.slack.com/archives/C023A7JDTJP/p1701366476953799). This one is due to Infura deployment, and they rolled back. This impacts mainnet, optimism, arbitrum, polygon.
2. Dec 12th [one](https://uniswapteam.slack.com/archives/C05021F43HU/p1702382483396269). Also a pure Infura fix. This impacts optimism, arbitrum.
3. Dec 18th [one](https://uniswapteam.slack.com/archives/C023A7JDTJP/p1702936311084879). This one I think arguably is an Infura issue, due to their node pool Bor upgrade, since we fallback to QuickNode and don't see the same issue (assuming quicknode didn't upgrade their node pool).

It's important to call out that we are interested in Infura issue not due to the underlying network issue. Because if it's the underlying network issue, then even if we switch to different provider, that chain is still under outage. I can think of:

1. Dec 15th [one](https://uniswapteam.slack.com/archives/C023A7JDTJP/p1702655386487909). Arbitrum sequencer outage. 
2. Aug 27th [one](https://uniswapteam.slack.com/archives/C023A7JDTJP/p1693205849389139). Polygon reorg, that caused missing trie node.

- Then we can ask ourselves, can we correlate the basic JSON RPC methods success rate to Infura issue but not underlying network issue?

1. `getBlockNumber()` does capture most of the cases, like Nov 30th and Dec 12th infura issue, meanwhile not falsefully classify Dec 15th one as pure infura issue:  Screenshot 2024-01-10 at 1 38 13 PM
2. `eth_call` would've misclassified arbitrum network issue as infura issue:
Screenshot 2024-01-10 at 1 46 41 PM
3. Nov 30th one is a broader Infura outage. We should not see quicknode outage on the same time:
Screenshot 2024-01-10 at 1 48 03 PM
Screenshot 2024-01-10 at 1 47 56 PM

- Finally, we need to ask ourselves, what do we do with the remaining but dominant RPC methods, like eth_call, when we need to roll out the eth_call health gauge?

The answer is also multifolded:
1. First we need to ensure we track eth_call at the highest level possible. Besides tracking under [UniJsonRpcProvider.ts](https://github.com/Uniswap/routing-api/pull/449/files#diff-28652acaf6b7750c3dc851cd9fb8328e66a1838af6e19c3fa55dc111689c2db5), we need to ensure [UniswapMulticallProvider](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/multicall-uniswap-provider.ts) all multicall methods are also tracked for health gauge.
2. How do we classify eth_call failures as Infura failure vs other issue? This is a tricky problem. I see you already noticed that eth_call throws 'PairLookupFailure'. On top level, I categorize 4 classes of issues:
i. Infura failure
ii. underlying network failure
iii. deterministic on-chain execution failures due to non-satisfying conditions (or 4xx in the server-side world). `PairLookupFailure` belongs to thi category.
iv. non-deterministic on-chain execution failures due to various issues, like header not found (or 5xx in the server-side world).

Unfortunately I don't have a golden answer here, and most likely in order to categorize eth_call, it requires repeated errors and trials there. For example, on-chain-quote-provider has coded some retryable [errors](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/on-chain-quote-provider.ts#L511-L550). Most likely the PerfStats will require us to hardcode those conditions as we go on.

I know this is a lot, but I hope this gives you some overview on how to think about the RPC failures due to different classes of issues, and how we can roll out gradually :)

lib/rpc/singleJsonRpcProvider.ts
```diff
@@ -0,0 +1,460 @@
...
+    }
+    debug(
+      `${this.url}: healthy: ${this.isHealthy()}, recovery ${timeInMs} * ${this.config.RECOVER_SCORE_PER_MS} = ${
+        timeInMs * this.config.RECOVER_SCORE_PER_MS
+      }, score => ${this.healthScore}`
+    )
+  }
+
+  private checkLastCallPerformance(method: string) {
+    if (!this.perf.lastCallSucceed) {
```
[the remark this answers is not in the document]
rollingtumbling: Thanks. Fixed as you suggested!

lib/rpc/GlobalRpcProviders.ts
```diff
@@ -0,0 +1,66 @@
...
+  private static initGlobalUniRpcProviders(log: Logger) {
+    if (this.SINGLE_RPC_PROVIDERS === null) {
+      this.initGlobalSingleRpcProviders(log)
+    }
+    this.UNI_RPC_PROVIDERS = new Map([
+      [
+        ChainId.MAINNET,
+        new UniJsonRpcProvider(
+          ChainId.MAINNET,
+          this.SINGLE_RPC_PROVIDERS!.get(ChainId.MAINNET)!,
```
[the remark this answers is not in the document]
rollingtumbling: Thanks. Added a check for that.

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -0,0 +1,235 @@
...
+  TransactionReceipt,
+  TransactionResponse,
+} from '@ethersproject/abstract-provider'
+import { BigNumber, BigNumberish } from '@ethersproject/bignumber'
+import { Deferrable } from '@ethersproject/properties'
+import { deriveProviderName } from '../handlers/evm/provider/ProviderName'
+import Logger from 'bunyan'
+
+interface SingleCallPerf {
+  isSucceed: boolean
```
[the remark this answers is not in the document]
rollingtumbling: Agreed. Fixed

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -0,0 +1,235 @@
...
+  latencyInMs: number
+  startTimestampInMs: number
+}
+
+export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
+  // TODO(jie): Implement block-aligned cache
```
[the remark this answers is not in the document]
rollingtumbling: Yes. It's a TODO item. But maybe can be moved to before the definition of this class.

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -0,0 +1,235 @@
...
+    this.getBlockNumber() // Ignore output in the promise
+  }
+
+  // Wrap another layer only for the sake of ease unit testing.
+  // We will test this API to represent the tests of other similar implemented APIs.
+  private _getBlockNumber(): Promise<number> {
+    return super.getBlockNumber()
+  }
+
+  private wrappedFunctionCall(fnName: string, fn: any, ...args: any[]): Promise<any> {
```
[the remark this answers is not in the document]
rollingtumbling: Are these two semantically identical? In my code there's no await, only callbacks.

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -0,0 +1,235 @@
...
+interface SingleCallPerf {
+  isSucceed: boolean
+  latencyInMs: number
+  startTimestampInMs: number
+}
+
+export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
+  // TODO(jie): Implement block-aligned cache
+  readonly url: string
+  readonly providerName: string
```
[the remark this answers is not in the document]
rollingtumbling: Yes, it's a sensible field to be read from this provider object.

lib/rpc/UniJsonRpcProvider.ts
```diff
@@ -0,0 +1,316 @@
...
+        this.checkUnhealthyProvider()
+      })
+  }
+
+  ///////////////////// Begin of override functions /////////////////////
+
+  // Notice: We should only intercept public methods that live at the top level and supposed to be called by user code
+
+  override getBlockNumber(sessionId?: string): Promise<number> {
+    return this.wrappedFunctionCall('getBlockNumber', sessionId)
```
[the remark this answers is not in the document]
rollingtumbling: Hmm...Do we minify code in this repo when deployment? Locally generated JS code doesn't seem to be minized.

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  private healthScore
+  private healthy: boolean
+  private lastCallTimestampInMs: number
+  private config: Config
+  private readonly metricPrefix: string
+  private readonly log: Logger
+
+  constructor(chainId: ChainId, url: string, log: Logger, config: Config = DEFAULT_CONFIG) {
+    super(url, { chainId, name: ID_TO_NETWORK_NAME(chainId) })
```
jsy1218, line 37: nit: the existing `network?: Networkish` within `InstrumentedEVMProviderProps` is more strongly-typed than the anonymous typed object `{ chainId, name: ID_TO_NETWORK_NAME(chainId) }` here. I would write as:

```
constructor(network: Networkish, url: string, log: Logger, config: Config = DEFAULT_CONFIG) {
    super(url, network)
    this.url = url
    this.log = log
    this.providerName = deriveProviderName(url)
    this.healthScore = 0
    this.healthy = true
    this.lastCallTimestampInMs = 0
    this.config = config
    this.metricPrefix = `RPC_${this.providerName}_${this.network.chainId}`
  }
```

meanwhile you don't need to change the caller code, because `Networkish` has a union type of number.

lib/rpc/SingleJsonRpcProvider.ts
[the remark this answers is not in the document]
jsy1218: I think Andy means:

```
private async wrappedFunctionCall(fnName: string, fn: any, ...args: any[]): Promise<any> {
    this.log.debug(`SingleJsonRpcProvider: wrappedFunctionCall: fnName: ${fnName}, fn: ${fn}, args: ${[...args]}`)
    const perf: SingleCallPerf = {
      succeed: true,
      latencyInMs: 0,
      startTimestampInMs: Date.now(),
    }

    try {
      return await fn(...args)
    } catch (error) {
      perf.succeed = false
      this.log.debug(JSON.stringify(error))
      throw error
    } finally {
      this.checkLastCallPerformance(fnName, perf)
      this.lastCallTimestampInMs = perf.startTimestampInMs
    }
  }
```

They are sementically equivalent, but try-catch-finally is a lot easier to read, as well as for future code maintenance perspective. We've been bitten by SOR functional programming style, where the functional programming made it more difficult to add finer granularity metrics, e.g. in SOR, we want to know the getQuotes only latencies, but the functional programming [implementation](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/alpha-router.ts#L1837-L1864) made it harder to add such metrics.

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -0,0 +1,235 @@
...
+      latencyInMs: 0,
+      startTimestampInMs: Date.now(),
+    }
+    return fn(...args)
+      .then((response: any) => {
+        return response
+      })
+      .catch((error: any) => {
+        perf.succeed = false
+        this.log.debug(JSON.stringify(error))
```
jsy1218, line 137: for logging error, it should be at error level. Note, any log level below INFO only has ~10% chance to get logged during routing-api runtime, because of this implementation https://github.com/Uniswap/routing-api/blob/main/lib/handlers/quote/injector.ts#L38.

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -0,0 +1,235 @@
...
+    this.getBlockNumber() // Ignore output in the promise
+  }
+
+  // Wrap another layer only for the sake of ease unit testing.
+  // We will test this API to represent the tests of other similar implemented APIs.
+  private _getBlockNumber(): Promise<number> {
+    return super.getBlockNumber()
+  }
+
+  private wrappedFunctionCall(fnName: string, fn: any, ...args: any[]): Promise<any> {
```
jsy1218, line 124: nit: I think `wrappedFunctionCall` can be slightly more strongly-typed:

`private async wrappedFunctionCall<ReturnType>(fnName: string, fn: (...args: any[]) => Promise<ReturnType>, ...args: any[]): Promise<ReturnType>`

lib/rpc/GlobalRpcProviders.ts
```diff
@@ -0,0 +1,69 @@
...
+      throw new Error(`No RPC providers configured for chain ${ChainId.MAINNET}`)
+    }
+    this.UNI_RPC_PROVIDERS = new Map([
+      [
+        ChainId.MAINNET,
+        new UniJsonRpcProvider(
+          ChainId.MAINNET,
+          this.SINGLE_RPC_PROVIDERS!.get(ChainId.MAINNET)!,
+          log,
+          GlobalRpcProviders.PROVIDER_RPC_URL_RANKING.get(ChainId.MAINNET),
```
jsy1218, line 54: don't forget to change to a non-mainnet before merging this PR. We don't have feature switch for provider gateway fallback, so it will take effect in prod immediately, upon merging the PR. Because by merging the PR, it will automatically deploy to staging and then prod (assuming the pipeline releases successfully)

lib/rpc/UniJsonRpcProvider.ts
```diff
@@ -0,0 +1,316 @@
...
+    // Dummy super constructor call is needed.
+    super('dummy_url', { chainId, name: 'dummy_network' })
+    this.connection.url
+    this.log = log
+
+    if (isEmpty(singleRpcProviders)) {
+      throw new Error('Empty singlePrcProviders')
+    }
+
+    if (ranking !== undefined && weights !== undefined && weights && ranking.length != weights.length) {
```
jsy1218, line 58: nit: `ranking?.length !== weights?.length`

It will also cover the 3 permutations of ranking is undefined and weights is undefined, which are all not going to throw error.

lib/rpc/UniJsonRpcProvider.ts
```diff
@@ -0,0 +1,316 @@
...
+    }
+
+    if (sessionId !== undefined && this.sessionCache.has(sessionId)) {
+      const provider = this.sessionCache.get(sessionId)!
+      if (provider.isHealthy()) {
+        return provider
+      } else if (!this.allowProviderAutoSwitch) {
+        throw new Error(
+          `Forced to use the same provider during the session but the provider (${provider.providerName}) is unhealthy`
+        )
```
jsy1218, line 99: I think I will log an error instead of throwing here. What if our heuristic is not good enough, and actually provider can service us?

jsy1218 (approved):

lib/rpc/UniJsonRpcProvider.ts
```diff
@@ -0,0 +1,316 @@
...
+
+  get lastUsedUrl() {
+    return this.lastUsedProvider?.url
+  }
+
+  disableFallback() {
+    this.totallyDisableFallback = true
+  }
+
+  createNewSessionId(): string {
```
cgkol, line 196: so who calls this? the rpcprovider before it makes the calls?

test/mocha/unit/rpc/UniJsonRpcProvider.test.ts
```diff
@@ -0,0 +1,687 @@
...
+    new SingleJsonRpcProvider(
+      ChainId.MAINNET,
+      `https://eth-mainnet.g.alchemy.com/v2/PC1uzrHueA8AdsD8jdQPcXFt4IUKSm-g`,
+      log
+    ),
+  ],
+}
+
+const resetRpcProviders = () => {
+  SINGLE_RPC_PROVIDERS[ChainId.MAINNET] = [
```
cgkol, line 46: same asline 27 right?

cgkol (approved):
Looks great Jie! only minor comments, mostly want an answer around the sessionId and how we are thinnking about integrating it

lib/rpc/GlobalRpcProviders.ts
cgkol: no code is actually reading from it so it should not matter right?

lib/handlers/injector-sor.ts
```diff
@@ -161,18 +162,23 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
...
-                  url: url,
-                  timeout,
-                },
-                network: chainId,
-                name: deriveProviderName(url),
-              }),
-            ],
-          }).getProvider()
+          let provider: StaticJsonRpcProvider
+          if (GlobalRpcProviders.getGlobalUniRpcProviders(log).has(chainId)) {
```
cgkol, line 166: we probably want to add a feature flag here instead of rolling it out immediately. WDYT?

cgkol (commented):
Yeah added one more point about rolling this one out

test/mocha/unit/rpc/UniJsonRpcProvider.test.ts
rollingtumbling: Thanks. Removed the duplication.

lib/handlers/injector-sor.ts
rollingtumbling: Now the flag feature has been added. Please see `app.ts` as well as `GlobalRpcProviders.ts`.

PR #457, opened 2024-02-03
chore: bump lambda version to route Polygon RPC provider from QuickNode back to Infura

jsy1218:
We want to route Polygon RPC provider from QuickNode back to Infura.


cgkol (approved):
👍 thank you siyu

jsy1218:
> 👍 thank you siyu

No problem. Will merge during EST Monday business hours.

PR #463, opened 2024-02-13
Shadow call backup healthy provider to collect and sync latency metrics

rollingtumbling:
This PR adds capability to
- Make shadow call for unselected but healthy providers to evaluate their latency and record it
- For selected provider, will also record latency from time to time
- Recorded latency will be synced to DB to propagate across different lambda instances
- Publish latency metric to cloudwatch

Some more details regarding syncing latency stat with DB:
- Regarding the latency stat sync with DB. We currently set to collect latency at most every 15 seconds. The DB sync happens every 5 seconds. So each provider only needs to hold a single latest latency evaluation result
- When syncing with the DB, latency records older than 5 minutes will be removed from DB

We also did some refactors
- Abstract the state, state diff and interface to interact with storage (repos)
- Make the state calculation (applying state diff) explicit and generalized enough to handle future state definitions
- Separate configs for `UniJsonRpcProvider` and `SingleJsonRpcProvider`

Currently the latency is only collected but not used to affect the provider selection logic.

UT and integration test passes. Manually tested on local dev environment by checking the logs and DB table items. For example, the DB table item will look like

```json
{
  "healthScore": 0,
  "latencies": [
    {
      "timestampInMs": 1707812500821,
      "latencyInMs": 66,
      "apiName": "getBlockNumber"
    },
    {
      "timestampInMs": 1707812460379,
      "latencyInMs": 130,
      "apiName": "getBlockNumber"
    }
  ]
}
```


rollingtumbling:
* `main`
  * **PR #463** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/463?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/463?utm_source=stack-comment).

PR #467, opened 2024-02-14
fix: two flaky tests

jsy1218:
BOND -> APE is failing in the pipeline. We already skip the same test in URA https://github.com/Uniswap/unified-routing-api/pull/351, this is despite some prior effort to attempt to make it more stable https://github.com/Uniswap/unified-routing-api/pull/349.

DAI -> WBTC test fails flakily, because the input DAI amount is too small. We replicate the same fix from https://github.com/Uniswap/unified-routing-api/pull/351.

Tested locally against prod endpoint: https://app.warp.dev/block/MHT4cW8PmLJWImHHtw5BsS


cgkol (approved):
✅

PR #468, opened 2024-02-15
Use separate envvar / AWS secret to contain all provider config

rollingtumbling:
Changes included
- Use a separate envvar / AWS secret to contain all provider's config (typed as `ProdConfig`), instead of relying on existing envvar and secrets. This will make it easier to manage the config for prod deployment
- Remove of some unnecessary features in `UniJsonRpcProvider`


lib/rpc/GlobalRpcProviders.ts
```diff
@@ -2,87 +2,89 @@ import { ChainId } from '@uniswap/sdk-core'
...
-    if (infuraAvalancheUrl === undefined) {
-      throw new Error(
-        `Infura Avalanche URL isn't provided by environment variable WEB3_RPC_${ChainId.AVALANCHE.toString()}`
-      )
+  private static getProdConfig(): ProdConfig {
+    const prodConfigStr = process.env['UNI_RPC_PROVIDER_PROD_CONFIG']!
+    if (prodConfigStr === undefined) {
+      throw new Error('Environment variable UNI_RPC_PROVIDER_PROD_CONFIG is missing!')
+    }
+    // Notice: No input validation! You need to be careful about the env var's format!
```
cgkol, line 24: this is very risky... shall we default to something then? if someone pushes the wrong enviornment then we can completely break it right?

lib/rpc/utils.ts
```diff
@@ -0,0 +1,24 @@
+import { ChainId } from '@uniswap/sdk-core'
+
+export function chainIdToNetworkName(networkId: ChainId): string {
+  switch (networkId) {
```
cgkol, line 4: does `SingleJsonRpcProvider` not accept the `ChainId` object? this seems redundant but tell me if i am missing something here?

cgkol (commented):
May i suggest we do first the deleting the unncessary fatures and then this one? both can break thee system so it would be good to separate

rollingtumbling:
@andysmith415 Comment addressed. Please take another look when you have a chance, thanks!

lib/rpc/GlobalRpcProviders.ts
rollingtumbling: > Thanks for adding this @rollingtumbling - one small point, historically we've ran either https://joi.dev/api/?v=17.12.0 for schema validation in the Routing API. Very similar to io-ts, just without the functional bend to it. Mind using that here as well? We've got a handful of examples here:
> 
> https://github.com/Uniswap/routing-api/blob/147f2204d6ee7ab78fe9a44fb48114707476e554/lib/handlers/quote/schema/quote-schema.ts#L22

Now use Joi instead. Please take another look!

rollingtumbling:
@andysmith415 Please take another look, thanks!

lib/rpc/GlobalRpcProviders.ts
```diff
@@ -2,87 +2,96 @@ import { ChainId } from '@uniswap/sdk-core'
...
-  private static initGlobalSingleRpcProviders(log: Logger, config: SingleJsonRpcProviderConfig) {
-    // Only Avalanche is supported for now.
-    const infuraAvalancheUrl = process.env[`WEB3_RPC_${ChainId.AVALANCHE.toString()}`]!
-    if (infuraAvalancheUrl === undefined) {
+  private static getProdConfig(): ProdConfig {
+    const prodConfigStr = process.env['UNI_RPC_PROVIDER_PROD_CONFIG']!
+    if (prodConfigStr === undefined) {
+      throw new Error('Environment variable UNI_RPC_PROVIDER_PROD_CONFIG is missing!')
+    }
+    const validation = ProdConfigJoi.validate(JSON.parse(prodConfigStr))
```
cgkol, line 24: do we want to fail or do we want to fallback to a default wdyt? if it fails then literally our routing-api goes down right?

rollingtumbling:
* **#475** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/475?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#469** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/469?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#468** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/468?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/468?utm_source=stack-comment).

PR #473, opened 2024-02-20
Prevent routing v2 on mobile

mikeki:
When selecting the protocols to route, we will check if the request source is for ios or android, and as long as the chainId is different than mainnet, then we avoid routing through those protocols.

Applying to both v2 and mixed routes (which also use v2)

### DIFF
--- a/lib/handlers/quote/quote.ts
+++ b/lib/handlers/quote/quote.ts
@@ -10,7 +10,7 @@ import {
   routeAmountsToString,
   SimulationStatus,
   SwapOptions,
-  SwapRoute
+  SwapRoute,
 } from '@uniswap/smart-order-router'
 import { Pool } from '@uniswap/v3-sdk'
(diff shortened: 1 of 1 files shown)


mattbspector (approved):

lib/handlers/quote/quote.ts
```diff
@@ -171,13 +171,23 @@ export class QuoteHandler extends APIGLambdaHandler<
       for (const protocolStr of protocolsStr) {
         switch (protocolStr.toLowerCase()) {
           case 'v2':
-            protocols.push(Protocol.V2)
+            if (
+              chainId === ChainId.MAINNET ||
+              !['uniswap-ios', 'uniswap-android'].includes(params.requestQueryParams.source ?? '')
```
tarikbellamine, line 176: nit: probably makes sense to separate these strings out into a `MOBILE_REQUEST_SOURCES` array or something

PR #483, opened 2024-02-23
fix: arbitrum use default multipler and integ-tests retry 3 times

jsy1218:
Wallet uses tenderly simulation. Default multipler is [1.3](https://github.com/Uniswap/smart-order-router/blob/main/src/providers/tenderly-simulation-provider.ts#L96). Routing-api overrides Arbitrum to 2.5, and this was to account for a bug on Tenderly back then after Nitro upgrade. Tenderly have fixed their side of bugs, so routing-api can remove 2.5 multipler. 

Also enable retry on integ-tests level.


cgkol (approved):
✅ good find

PR #484, opened 2024-02-23
Use config file for RPC Provider Config

rollingtumbling:
This is a reworked version of https://github.com/Uniswap/routing-api/pull/468.

Compared with the original PR, instead of using AWS secret (eventually becomes environmental variables), we now use separate config file.

We only use environmental variables for provider URLs. This is to minimize the usage of AWS secrets which turns out to be a pain to update and test.

Tested on local env and unit test.


rollingtumbling:
* **#484** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/484?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/484?utm_source=stack-comment).

cgkol (approved):
this looks good; this is tested e2e right?

rollingtumbling:
> this looks good; this is tested e2e right?

Yes, tested on my local env.

PR #486, opened 2024-02-26
Add SEPOLIA to V2Supported

mikeki:
Forgot to add Sepolia in https://github.com/Uniswap/routing-api/pull/488


jsy1218:
GOERLI is not deployed with universal router and v2 factory, it won't work right?

jsy1218 (approved):

PR #492, opened 2024-03-05
chore: remove sepolia from v2 routing (#491)

rollingtumbling:
chore: remove sepolia from v2 routing (#491)

remove sepolia from v2 routing

[RPC Gateway] Force to use the same provider when serving each quote request

This PR will forbid provider switch during serving the same quote request.
The selection of provider will still happen when serving a new quote request.


rollingtumbling:
* **#492** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/492?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

lib/handlers/quote/quote.ts
```diff
@@ -57,6 +58,13 @@ export class QuoteHandler extends APIGLambdaHandler<
     let result: Response<QuoteResponse> | ErrorResponse
 
     try {
+      for (const chainId of SUPPORTED_CHAINS) {
+        if (GlobalRpcProviders.getGlobalUniRpcProviders(log).has(chainId)) {
+          const provider = GlobalRpcProviders.getGlobalUniRpcProviders(log).get(chainId)!
+          provider.forceAttachToNewSession()
+        }
+      }
+
```
mikeki, line 67: I think we should be able to use the `chainId` defined in line 47, since a given request will only execute for a single chain.

mikeki (changes requested):
Adding the request for changes, before we merge. I think the way you are doing it right now is initializing the session for all SUPPORTED_CHAINS but we only need it on the request chainId

mikeki (approved):
Thanks :)

cgkol (approved):
seems reasonable!

PR #517, opened 2024-03-16
Lambda version bump

mikeki:


jsy1218:
curious why we skip version 13? I might have to use this PR, because we have a new blast URL now.

jsy1218 (approved):

PR #518, opened 2024-03-16
Launch RPC gateway to 6 chains

rollingtumbling:
This launches RPC gateway to Avalance, Optimsm, Celo, BNB, Polygon and Base.

Testing
- Deploy to local env works
- Sample requests (for these 6 chains) to local env works
- Local env dashboard looks fine
- Didn't encounter env variable 4KB limit issue when deploying to local env

### DIFF
--- a/bin/app.ts
+++ b/bin/app.ts
@@ -12,6 +12,7 @@ import 'source-map-support/register'
 import { SUPPORTED_CHAINS } from '../lib/handlers/injector-sor'
 import { STAGE } from '../lib/util/stage'
 import { RoutingAPIStack } from './stacks/routing-api-stack'
+import { getRpcGatewayEnabledChainIds } from '../lib/rpc/ProdConfig'
 
 dotenv.config()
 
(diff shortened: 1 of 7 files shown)


rollingtumbling:
* **#518** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/518?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

bin/app.ts
```diff
@@ -159,16 +160,54 @@ export class RoutingAPIPipeline extends Stack {
...
+      'QUICKNODE_137',
+      'ALCHEMY_137',
+      // Base
+      'INFURA_8453',
+      'QUICKNODE_8453',
+      'ALCHEMY_8453',
+      'NIRVANA_8453',
+    ]
+    for (const provider of RPC_GATEWAY_PROVIDERS) {
+      jsonRpcProviders[provider] = jsonRpcProvidersSecret.secretValueFromJson(provider).toString()
```
mikeki, line 205: I think this could take us over the 4kb limit.

Is there any way we could simplify the secrets? Like all we need for Infura is an API key, every URL can be hardcoded based on the chain, I think there are some providers that return random urls, but it would be helpful to figure out if there's a way to simplify this, we are essentially multiplying our providers for every chain which is not gonna work as we need to expand to new chains.

jsy1218:
We should delete some existing secrets before we roll out more new JSON RPCs. For example, Goerli is deprecated in routing-api, so `JSON_RPC_PROVIDER_5` can be safely deleted.

bin/app.ts
jsy1218: I had the same suggestion https://uniswapteam.slack.com/archives/C021SU4PMR7/p1710359540530009?thread_ts=1710353824.086039&cid=C021SU4PMR7

jsy1218: Especially with rollup-centric road and rollup-as-a-service offering, we will likely roll out to new chain at faster speed.

rollingtumbling: @mikeki Made some changes to reduce the env var's length. Please take a look!

rollingtumbling:
> We should delete some existing secrets before we roll out more new JSON RPCs. For example, Goerli is deprecated in routing-api, so `JSON_RPC_PROVIDER_5` can be safely deleted.
> 
> Once we delete the obsolete secrets, we can have a dedicated PR that just bumps the lambda version, and ensures it still works.

Secrets size doesn't count against 4KB limit. Also I would suggest we keep old secret because of what happened last week...Removing the code doesn't mean the old secrets aren't needed...

mikeki (approved):
Do not merge yet.

We still need to:
1. Update the lambda version to guarantee the new secrets are loaded
2. Update the secrets

rollingtumbling:
> Do not merge yet.
> 
> We still need to:
> 
> 1. Update the lambda version to guarantee the new secrets are loaded
> 2. Update the secrets

1. Version has been updated.
2. Need your help update the secrets! Thanks

PR #524, opened 2024-03-20
feat: shadow quote new quoter v2 on Polygon Mumbai and Sepolia

jsy1218:
We want to switch to new view-only quoter. But we are not ready to live migrate yet. In this PR, we want to achieve:

1. Shadow request very small percent of Polygon quote against the new quoter v2, without impacting the routing-api endpoint success rate at all.
2. Make small refactoring in this PR, as well as based on https://github.com/Uniswap/smart-order-router/pull/510, so that it's easy to make incremental changes, when we roll out quoter to
   - same chain at incremental live traffic migration to new quoter v2
   - to new chain with the shadow quote, meanwhile using the correct quote config params without the manual copy-paste errors
3. Collect sampling metrics against the shadow new quoter v2, so as to understand
   - is new quoter faster than the existing quoter
   - is new quoter way gas-efficient than existing quoter
   - is new quoter retrying less frequent than exising quoter.

For Polygon, existing quoter batch size is about 200k/minute, so we choose 0.1% to have additional 200 batched inputs/minute:

Screenshot 2024-03-21 at 4 52 01 PM

TODO:

- [x] check polygon quoter batch size metrics, so as to determine the actual sampling percent
- [x] refactor in SOR to have default configs, so that they can be reused by routing-api
- [x] add unit tests against traffic switch
- [x] merge https://github.com/Uniswap/smart-order-router/pull/510, so that this PR can compile. Some of this PR changes is assumed from https://github.com/Uniswap/smart-order-router/pull/510.

### DIFF
--- a/lib/handlers/quote/provider-migration/v3/traffic-switch-on-chain-quote-provider.ts
+++ b/lib/handlers/quote/provider-migration/v3/traffic-switch-on-chain-quote-provider.ts
@@ -14,10 +14,14 @@ export class TrafficSwitchOnChainQuoteProvider implements IOnChainQuoteProvider 
   private readonly currentQuoteProvider: IOnChainQuoteProvider
   private readonly targetQuoteProvider: IOnChainQuoteProvider
 
-  protected readonly SHOULD_SWITCH_TRAFFIC = () =>
-    QUOTE_PROVIDER_TRAFFIC_SWITCH_CONFIGURATION.switchPercentage > this.getRandomPercentage()
-  protected readonly SHOULD_SAMPLE_TRAFFIC = () =>
-    QUOTE_PROVIDER_TRAFFIC_SWITCH_CONFIGURATION.samplingPercentage > this.getRandomPercentage()
+  protected readonly SHOULD_SWITCH_EXACT_IN_TRAFFIC = () =>
+    QUOTE_PROVIDER_TRAFFIC_SWITCH_CONFIGURATION.switchExactInPercentage > this.getRandomPercentage()
+  protected readonly SHOULD_SAMPLE_EXACT_IN_TRAFFIC = () =>
+    QUOTE_PROVIDER_TRAFFIC_SWITCH_CONFIGURATION.samplingExactInPercentage > this.getRandomPercentage()
+  protected readonly SHOULD_SWITCH_EXACT_OUT_TRAFFIC = () =>
+    QUOTE_PROVIDER_TRAFFIC_SWITCH_CONFIGURATION.switchExactOutPercentage > this.getRandomPercentage()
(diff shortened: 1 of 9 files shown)


lib/handlers/quote/provider-migration/v3/traffic-switch-on-chain-quote-provider.ts
```diff
@@ -0,0 +1,102 @@
...
+    let currentQuote
+    let targetQuote
+
+    metric.putMetric('ON_CHAIN_QUOTE_PROVIDER_EXACT_IN_TRAFFIC_TOTAL', 1, MetricLoggerUnit.None)
+
+    if (sampleTraffic) {
+      metric.putMetric('ON_CHAIN_QUOTE_PROVIDER_EXACT_IN_TRAFFIC_SAMPLING', 1, MetricLoggerUnit.None)
+
+      currentQuote = await this.currentQuoteProvider.getQuotesManyExactIn(amountIns, routes, providerConfig)
+      targetQuote = await this.targetQuoteProvider.getQuotesManyExactIn(amountIns, routes, providerConfig)
```
mikeki, line 51: We should send these 2 in parallel right?

mikeki: Do we want to collect some latency metrics?

lib/handlers/quote/provider-migration/v3/traffic-switch-on-chain-quote-provider.ts
```diff
@@ -0,0 +1,102 @@
...
+    let currentQuote
+    let targetQuote
+
+    metric.putMetric('ON_CHAIN_QUOTE_PROVIDER_EXACT_OUT_TRAFFIC_TOTAL', 1, MetricLoggerUnit.None)
+
+    if (sampleTraffic) {
+      metric.putMetric('ON_CHAIN_QUOTE_PROVIDER_EXACT_OUT_TRAFFIC_SAMPLING', 1, MetricLoggerUnit.None)
+
+      currentQuote = await this.currentQuoteProvider.getQuotesManyExactOut(amountOuts, routes, providerConfig)
+      targetQuote = await this.targetQuoteProvider.getQuotesManyExactOut(amountOuts, routes, providerConfig)
```
mikeki, line 85: ditto

lib/handlers/quote/provider-migration/v3/traffic-switch-on-chain-quote-provider.ts
jsy1218: I parallelized those two calls. There's also a nuance now, if the target quoter calls throws runtime exception during sampling, we will want to swallow the exception, because otherwise sampling traffic is no longer shadow traffic.

PR #528, opened 2024-03-22
chore: per-chain alarm and per-request-source alarm

jsy1218:
For per-chain alarm, we already coded it up. Sadly it was sourcing from incorrect place, i.e. API Gateway, that's why the per-chain alarm never goes off in prod:
Screenshot 2024-03-22 at 2 25 04 PM

I changed it to source from Uniswap/RoutingAPI, since we log the metrics from the routing lambda, and verified from my local the alarm is working:

Screenshot 2024-03-22 at 2 25 01 PM

For per-request-source alarm, I noticed we even have undefined in our prod traffic, which is not trivial:
Screenshot 2024-03-22 at 1 59 22 PM

So the per-request-source alarm will begin with every possible request source:
Screenshot 2024-03-22 at 2 25 11 PM


lib/util/requestSources.ts
```diff
@@ -0,0 +1,9 @@
+export const REQUEST_SOURCES = [
```
cgkol, line 1: maybe we should put these in an SDK one day?

bin/stacks/routing-api-stack.ts
```diff
@@ -374,16 +381,25 @@ export class RoutingAPIStack extends cdk.Stack {
       const metric = new MathExpression({
         expression: '100*(response200/(invocations-response400))',
         usingMetrics: {
-          invocations: api.metric(`GET_QUOTE_REQUESTED_CHAINID: ${chainId.toString()}`, {
-            period: Duration.minutes(5),
+          invocations: new aws_cloudwatch.Metric({
+            namespace: 'Uniswap',
+            metricName: `GET_QUOTE_REQUESTED_CHAINID: ${chainId.toString()}`,
```
cgkol, line 386: so we already had per chain? why was it not working?

cgkol (commented):
Siyu, i think it might be best if we do per source per chain instead of just source? or maybe both? because its important to know if one of the clients its only sending us bad stuff for one chain (because if tis wallet on base for example it will always be hidden because web with more calls wont surface it) 

WDYT?

lib/util/requestSources.ts
jsy1218: agree. URA is missing couple sources, and would benefit from shared SDK

bin/stacks/routing-api-stack.ts
jsy1218: Because it was incorrectly implemented. per-chain metrics are logged from routing lambda, but the implementation was trying to get the per-chain metrics from API gateway, so that the alarm always has insufficient datapoints to evaluate, resulting in no alarm across all time.

I fixed it here, and verified now it's working on my local.

mikeki: I think this is just the metric but not the alert, the alert is whats added on line 417. That's what I understand

mikeki: I originally thought this was just the metric, but I saw then that it is indeed the alert.

jsy1218: nah 5 minutes evaluation period is fine

jsy1218:
> Siyu, i think it might be best if we do per source per chain instead of just source? or maybe both? because its important to know if one of the clients its only sending us bad stuff for one chain (because if tis wallet on base for example it will always be hidden because web with more calls wont surface it)
> 
> WDYT?

I thought about per-chain per-source alarm, but I feel in the incident like last Tuesday, there might be too many alarms to oncall to ack at the same time, if multiple chains were affected. @mikeki Curious to see your thoughts.

jsy1218:
> > > Siyu, i think it might be best if we do per source per chain instead of just source? or maybe both? because its important to know if one of the clients its only sending us bad stuff for one chain (because if tis wallet on base for example it will always be hidden because web with more calls wont surface it)
> > > WDYT?
> > 
> > I thought about per-chain per-source alarm, but I feel in the incident like last Tuesday, there might be too many alarms to oncall to ack at the same time, if multiple chains were affected. @mikeki Curious to see your thoughts.
> 
> Yeah per-chain per-source feels like it will become very noisy, but we could consider keeping it without triggering an alert? 

That's a good idea, we can implement the alarm, but make the actionEnabled=false, so that it never triggers.

cgkol:
what if we do only sev3? @mikeki @jsy1218 wdyt? btw if its noisy thats a good thing still i would think if something is broken we should be alerting right? am i wrong? if its real noise of failrures than we should know

jsy1218:
> what if we do only sev3? @mikeki @jsy1218 wdyt? btw if its noisy thats a good thing still i would think if something is broken we should be alerting right? am i wrong? if its real noise of failrures than we should know

yea we could initially roll out this way. I'm keen to see if there will be per-source per-chain only alarm in prod.

mikeki (approved):

PR #529, opened 2024-03-23
Fix issues in existing shadow call & DB sync logic.

rollingtumbling:
Fix an issue where we didn't make shadow latency evaluation call with the correct `CallType`. This issue can lead to some of the measured methods failed to generate latency evaluation result.

Fix an issue where we started too many DB sync request inside a single lambda. This issue can lead to more DynamoDB reads and optimistic-write failures than expected.

Both issue fixes have been tested with RPC gateway config for various chains in local env.

Merging this PR won't impact production serving.


rollingtumbling:
* **#523** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/523?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#529** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/529?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -332,6 +335,8 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
     } catch (err: any) {
       this.log.error(`Encountered unhandled error when sync provider state: ${JSON.stringify(err)}`)
       // Won't throw. A fail of sync won't affect how we do health state update locally.
+    } finally {
+      this.syncingDb = false
```
cgkol, line 339: this is proabbly gonna work, but we keep running into this issue where we have boolean variables that are turned on in one place and you have to turn them on in many other places (similar to "doing helath check" which was your previous PR. Is there a safer way to do this potentially?

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -360,46 +365,40 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
 
   ///////////////////// Begin of override functions /////////////////////
 
-  override getBlockNumber(callType = CallType.NORMAL): Promise<number> {
-    return this.wrappedFunctionCall(callType, 'getBlockNumber', this._getBlockNumber.bind(this))
+  override getBlockNumber(): Promise<number> {
+    return this.wrappedFunctionCall(CallType.NORMAL, 'getBlockNumber', this._getBlockNumber.bind(this))
```
cgkol, line 369: why did we change this to always normal?

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -472,14 +465,41 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
...
+    return this.wrappedFunctionCall(CallType.NORMAL, 'call', super.call.bind(this), transaction, blockTag)
+  }
+
+  override send(method: string, params: Array<any>): Promise<any> {
+    return this.wrappedFunctionCall(CallType.NORMAL, 'send', super.send.bind(this), method, params)
+  }
+
+  ///////////////////// Begin of special functions /////////////////////
+
+  // Wrap another layer only for the sake of ease unit testing.
```
cgkol, line 478: i dont see any unit tests for this one so not sure why we created this functions? Also why is it easier to unit test i am a little confused here, we are adding boilerplate code is my undersatnding right?

jsy1218:
> Fix an issue where we started too many DB sync request inside a single lambda. This issue can lead to more DynamoDB reads and optimistic-write failures than expected.

Can you share the metrics on what's the current DynamoDB reads and writes request volumes? And how much this PR can reduce to?

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -296,13 +295,17 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
       return await fn(...args)
     } catch (error: any) {
       perf.succeed = false
-      this.log.error(JSON.stringify(error))
+      this.log.debug(
```
jsy1218, line 298: can't you just merge this debugging logging with the below error logging?

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -296,13 +295,17 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
...
+        `Provider call failed: provider: ${this.url}, fnName: ${fnName}, fn: ${fn}, args: ${JSON.stringify([...args])}`
+      )
+      this.log.error(`Error details: ${JSON.stringify(error)}`)
       throw error
     } finally {
       perf.latencyInMs = Date.now() - perf.startTimestampInMs
       this.checkLastCallPerformance(perf)
       this.updateHealthyStatus()
-      if (this.enableDbSync) {
+      if (this.enableDbSync && !this.syncingDb) {
```
jsy1218, line 307: `enableDbSyn` and `syncingDb` feel like duplicate configs to me. `syncingDb` hardcoded to false feels like a shortcut to me, that may potentially get lost in future.

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -229,18 +230,18 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
     this.log.debug(`${this.url}: Evaluate healthiness for unhealthy provider...`)
     this.evaluatingHealthiness = true
     try {
-      await this.getBlockNumber(CallType.HEALTH_CHECK)
+      await this.getBlockNumber_EvaluateHealthiness()
```
jsy1218, line 233: I'm confused on why you need to introduce another layer, meanwhile the overriden RPC methods like `getBlockNumber` can accept the optional param `CallType` to distinguish between normal check vs others?

lib/rpc/SingleJsonRpcProvider.ts
rollingtumbling: No they are different. `enableDbSync` means whether we enabled DB sync feature at all and it's only set in the constructor. `syncingDb` on the other hand is set to indicate we are in the progress of a DB sync (act like a lock to prevent future sync requests from being executed).

lib/rpc/SingleJsonRpcProvider.ts
rollingtumbling: Since they have to be set at the beginning and the end of an operation (DB sync, evaluate latency, evaluate health), it's hard to combine both access into one.

One thing I could slightly improve the robustness is that,  I will ignore the check for the flag if we haven't recorded the metrics twice long as expected. So even if we made some mistake, we don't deadlock forever. Please see my changes in `checkUnhealthyProviders()` and `checkOtherHealthyProvider()` in `UniJsonRpcProvider.ts`.

lib/rpc/SingleJsonRpcProvider.ts
rollingtumbling: Because for `HEALTH_CHECK` (in `getBlockNumber_EvaluateHealthiness`) and `LATENCY_EVALUATION` calls (in `getBlockNumber_EvaluateLatency`), we will use `this._getBlockNumber` instead of `getBlockNumber`. So this `getBlockNumber`'s only use case is to make normal calls.

lib/rpc/SingleJsonRpcProvider.ts
rollingtumbling: It's certainly not adding boilerplate code. The reason we have a `_getBlockNumber()` besides `getBlockNumber()` is that we want to mock the output (such as provider network call response) without removing the call of `wrappedFunctionCall()`, which contains important logic that a lot of UT cases will check against. Yes the only purpose is to make the mocking easier for UT.

I don't want to do this for other APIs since it just pollute the code too much. So I only did it for `getBlockerNumber()` API. It's good enough to represent major use cases though.

lib/rpc/SingleJsonRpcProvider.ts
rollingtumbling: I didn't add the metric for log DB sync. I realized that I made more DB sync than necessary is based on spotting the error of optimistic write failure, which indicates two write-after-read operations happens too close together. I just updated the PR to add DB sync metric reporting.

rollingtumbling:
> > Fix an issue where we started too many DB sync request inside a single lambda. This issue can lead to more DynamoDB reads and optimistic-write failures than expected.
> 
> Can you share the metrics on what's the current DynamoDB reads and writes request volumes? And how much this PR can reduce to?

I didn't add the metric for log DB sync. I realized that I made more DB sync than necessary is based on spotting the error of optimistic write failure, which indicates two write-after-read operations happens too close together. I just updated the PR to add DB sync metric reporting.

lib/rpc/SingleJsonRpcProvider.ts
uni-guillaume: it looks like you are trying to implement an optimistic lock with boolean, which probably won't work because your if condition is not atomic. Thoughts on either using an existing library or implementing our own so we can transform this block into the following ? 
```
if (this.enableDbSync){
   this.lock.tryAcquire(this.maybeSyncAndUpdateProviderState)
}
```

IMHO, it would clean up a bit the booleans in this class and provide this primitive for other projects

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -272,10 +273,12 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
     metric.putMetric(`${this.metricPrefix}_selected`, 1, MetricLoggerUnit.Count)
   }
 
-  // Wrap another layer only for the sake of ease unit testing.
-  // We will test this API to represent the tests of other similar implemented APIs.
-  private _getBlockNumber(): Promise<number> {
-    return super.getBlockNumber()
+  logDbSync(success: boolean) {
```
uni-guillaume, line 276: nit: since this is mostly syntactic sugar: thoughts on creating two function instead of one : 
```
logSuccessDbSync(){ metric.putMetric(`${this.metricPrefix}_db_sync_success`, 1, MetricLoggerUnit.Count) }
logFailureDbSync(){ metric.putMetric(`${this.metricPrefix}_db_sync_failure`, 1, MetricLoggerUnit.Count) }
```

This would save some navigation back and forth for the reader

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -472,14 +474,41 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
...
+  ///////////////////// Begin of special functions /////////////////////
+
+  // Wrap another layer only for the sake of ease unit testing.
+  // We will test this API to represent the tests of other similar implemented APIs.
+  private _getBlockNumber(): Promise<number> {
+    return super.getBlockNumber()
+  }
+
+  private getBlockNumber_EvaluateHealthiness(): Promise<number> {
+    return this.wrappedFunctionCall(CallType.HEALTH_CHECK, 'getBlockNumber', this._getBlockNumber.bind(this))
```
uni-guillaume, line 494: i'm curious about this pattern: is the binding necessary since `_getBlockNumber` is already a method of this class ?

rollingtumbling: This is necessary as we are forwarding the function to another function `wrappedFunctionCall`. Without this bind, it will use the wrong `this` context.

rollingtumbling: I tried it to verify that it is indeed needed..

lib/rpc/SingleJsonRpcProvider.ts
cgkol: `if we haven't recorded the metrics twice long as expected` -> what does this mean exactly btw?

jsy1218 (approved):
Im hoping this PR can mitigate the routing-api beta lambda cold start issue.

lib/rpc/SingleJsonRpcProvider.ts
rollingtumbling: > `if we haven't recorded the metrics twice long as expected` -> what does this mean exactly btw?

For example, we should wait for the lock if it's already 15 seconds since last healthiness check. However, just in case the lock was failed to release, we will allow a re-acquire of the lock 30 seconds after last healthiness check. This should actually happen, but the logic is here to prevent stalling. Same logic for doing latency check shadow calls.

PR #532, opened 2024-03-27
fix: only sev3 on testnets and unknown source

jsy1218:
We had couple pages on unknown source and testnets at sev-2. Those are not as business critical at sev-2, so only alert them at sev-3.


bin/stacks/routing-api-stack.ts
```diff
@@ -458,10 +463,6 @@ export class RoutingAPIStack extends cdk.Stack {
     const successRateByRequestSourceAndChainIdAlarm: cdk.aws_cloudwatch.Alarm[] = []
     REQUEST_SOURCES.forEach((requestSource) => {
       SUPPORTED_CHAINS.forEach((chainId) => {
-        if (CHAINS_NOT_MONITORED.includes(chainId)) {
-          return
-        }
-
```
mikeki, line 464: should we keep this?

lib/util/testNets.ts
```diff
@@ -0,0 +1,15 @@
...
+  ChainId.GOERLI,
+  ChainId.SEPOLIA,
+  ChainId.CELO_ALFAJORES,
+  ChainId.BASE_GOERLI,
+  ChainId.OPTIMISM_SEPOLIA,
+  ChainId.OPTIMISM_GOERLI,
+  ChainId.ARBITRUM_SEPOLIA,
+  ChainId.ARBITRUM_GOERLI,
+  ChainId.ZORA_SEPOLIA,
+]
```
rollingtumbling, line 15: New line at the bottom.

bin/stacks/routing-api-stack.ts
jsy1218: I think we should remove it, because per-source per-chain alarm is at sev-3, and we should get alerted on testnets with sev-3. We can adjust accordingly, if we find it too noisy.

bin/stacks/routing-api-stack.ts
```diff
@@ -18,8 +18,10 @@ import { RoutingLambdaStack } from './routing-lambda-stack'
 import { RoutingDatabaseStack } from './routing-database-stack'
 import { RpcGatewayDashboardStack } from './rpc-gateway-dashboard'
 import { REQUEST_SOURCES } from '../../lib/util/requestSources'
+import { TESTNETS } from '../../lib/util/testNets'
 
-export const CHAINS_NOT_MONITORED: ChainId[] = [ChainId.GOERLI, ChainId.POLYGON_MUMBAI]
+export const CHAINS_NOT_MONITORED: ChainId[] = TESTNETS;
```
rollingtumbling, line 23: Are we still using this const?

mikeki (approved):

bin/stacks/routing-api-stack.ts
jsy1218: yep, in https://github.com/Uniswap/routing-api/blob/main/bin/stacks/routing-api-stack.ts#L342 and  https://github.com/Uniswap/routing-api/blob/main/bin/stacks/routing-api-stack.ts#L377

PR #537, opened 2024-03-29
chore: do not alert testnet and unknown source on sev-3

jsy1218:
We have too many noisy pages over the last couple days, flooding with the important alerts. To begin mitigate, we can stop alerting on testnets and unknown source together.


rollingtumbling (approved):
Thanks Siyu!

PR #538, opened 2024-03-29
[RPC Gateway] Optimize DB sync rate limit logic, add more metrics & dashboard graph

rollingtumbling:
The logic DB sync rate limit has been improved to make the ownership much clearer between `SingleJsonRpcProvider` and `ProviderStateSyncer`. Previously the rate limit logic spanned across these two classes but now it's all in `SingleJsonRpcProvider`. Vague words like `maybe` are also removed to make the code easier to follow.

Added more unit test for checking the effect of DB sync rate limit.

More Cloudwatch metrics and dashboard graphs have also been added to verify the effect of local env testing and it will be very useful for production deployment monitoring.

Tested and verified in local env.

### DIFF
--- a/lib/rpc/SingleJsonRpcProvider.ts
+++ b/lib/rpc/SingleJsonRpcProvider.ts
@@ -228,6 +228,7 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
 
   async evaluateHealthiness() {
     this.log.debug(`${this.url}: Evaluate healthiness for unhealthy provider...`)
+    this.logCheckHealth()
     this.evaluatingHealthiness = true
     try {
       await this.getBlockNumber_EvaluateHealthiness()
@@ -239,6 +240,7 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
 
   async evaluateLatency(methodName: string, ...args: any[]) {
     this.log.debug(`${this.url}: Evaluate for latency... methodName: ${methodName}`)
+    this.logEvaluateLatency()
     this.evaluatingLatency = true
     try {
       await (this as any)[`${methodName}_EvaluateLatency`](...args)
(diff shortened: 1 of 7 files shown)


rollingtumbling:
* **#538** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/538?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -313,18 +332,23 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
       perf.latencyInMs = Date.now() - perf.startTimestampInMs
       this.checkLastCallPerformance(perf)
       this.updateHealthyStatus()
-      if (this.enableDbSync && !this.syncingDb) {
-        this.syncingDb = true
-        // Fire and forget. Won't check the sync result.
-        this.maybeSyncAndUpdateProviderState()
+      if (this.enableDbSync) {
+        if (
+          (!this.syncingDb && this.hasEnoughWaitSinceLastDbSync(1000 * this.config.DB_SYNC_INTERVAL_IN_S)) ||
```
jsy1218, line 337: do we really still need `syncingDb`? Second boolean condition evaluates to true just purely based on the time lag.

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -341,6 +365,8 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
         // Update latency stat
         this.updateLatencyStat(newState)
       }
+      this.lastDbSyncTimestampInMs = Date.now()
+      this.log.debug('Successfully synced with DB and updated states')
```
jsy1218, line 369: nit: you can put log debug into `logDbSyncSuccess()`, although the debug statement itself seems unnecessary because you are logging the success metric.

jsy1218 (approved):
minor comments

lib/rpc/SingleJsonRpcProvider.ts
rollingtumbling: You are right. I think I'm being too protective here at the cost of DB access flooding when creating a new lambda (which isn't good). I removed this logic.

lib/rpc/SingleJsonRpcProvider.ts
rollingtumbling: The logging is useful when looking through the logs after setting the logging level to `INFO`. I want to make the `logDbSyncSuccess()` specific to metric reporting. So having this seemingly extra two calls is fine I guess.

PR #542, opened 2024-04-02
[RPC gateway] Fully launch on Celo

rollingtumbling:
10% of Celo's traffic have been on RPC gateway for a day and the latency and success rate looks normal.

Fully launch won't increase the pressure to the major provider for Celo (Quicknode), but will increase the traffic for shadow provider (Infura). However we've seen very little shadow call made to Infura in the last few days after launch.

The QPS is about 0.1. So it won't add much pressure to Infura even if we turn it to full.

PR #543, opened 2024-04-02
[RPC gateway] Launch on 10% of Avalanche

rollingtumbling:
Since we launch 1% to Avalanche yesterday (4/1) , we don't see major metrics diff.

Now we want to slightly add some traffic to Avalanche. From 1% to 10%.

### DIFF
--- a/lib/config/rpcProviderProdConfig.json
+++ b/lib/config/rpcProviderProdConfig.json
@@ -7,7 +7,7 @@
   },
   {
     "chainId": 43114,
-    "useMultiProviderProb": 1,
+    "useMultiProviderProb": 0.1,
     "providerInitialWeights": [1, 0, 0],
     "providerUrls": ["INFURA_43114", "QUICKNODE_43114", "NIRVANA_43114"]
   }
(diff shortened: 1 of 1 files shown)


jsy1218 (commented):
we need the same rigor as https://github.com/Uniswap/routing-api/pull/542 for the [comment](https://github.com/Uniswap/routing-api/pull/542#issue-2221202394) of the RPC rollout metrics so far.

lib/config/rpcProviderProdConfig.json
```diff
@@ -1,13 +1,13 @@
 [
   {
     "chainId": 42220,
-    "useMultiProviderProb": 0.1,
+    "useMultiProviderProb": 1,
```
jsy1218, line 4: I think you need to update your branch, as this was from https://github.com/Uniswap/routing-api/pull/542

jsy1218 (approved):

lib/config/rpcProviderProdConfig.json
rollingtumbling: Yes. Added.

PR #544, opened 2024-04-02
fix: filter out expired routes by ttl in the RoutesDB read path

jsy1218:
In prod, we noticed that the RoutesDB can return cached routes, that we can no longer query from RoutesDB:

Screenshot 2024-04-01 at 3 49 39 PM

From https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html, "You can still update the expired items that are pending deletion, including changing or removing their TTL attributes". This can explain why the expired routes can still get returned from RoutesDB and used in router for quoting.

Testing wise, I used e2e tests on my local https://app.warp.dev/block/zcIboiQ5gWWjNGat5cWsFE. The idea here is that, meanwhile the tests are running, I expect the RoutesDB to get queried live one by one, so [RoutesDbQuery](https://github.com/Uniswap/routing-api/blob/c71d31bd040b18186c3d87e5692000327538d77f/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L150) is the highest. Then I expect [RoutesDbPreFilterEntriesFound](https://github.com/Uniswap/routing-api/blob/c71d31bd040b18186c3d87e5692000327538d77f/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L166) and [RoutesDbEntriesFound](https://github.com/Uniswap/routing-api/blob/c71d31bd040b18186c3d87e5692000327538d77f/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L199C23-L199C43) to be always equal, and the reason is because we just created the routes in the local RoutesDB, there's no way they can get filtered out by unmatched protocol version, or TTL expiry:


rollingtumbling (approved):
When the record get written to DB, does it have TTL set already?

jsy1218:
> When the record get written to DB, does it have TTL set already?

That's right. It's written [here](https://github.com/Uniswap/routing-api/blob/c71d31bd040b18186c3d87e5692000327538d77f/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L394).

mikeki (approved):

cgkol (commented):
Can we add a test please?

cgkol (changes requested):
let me change this

cgkol:
@rollingtumbling @mikeki reminder to make sure we enforce tests especially on change that affect production

jsy1218:
> Can we add a test please?

Added an integ-test, that actually cached routes into the RoutesDB dynamo DB, and then use the `mockdate` library to make sure the system time shifts to one later later, so that retrieving the cached routes can be actually filtered due to the TTL expiry.

cgkol (commented):
Lets make sure @mikeki gets to see it too!

jsy1218:
> Lets make sure @mikeki gets to see it too!
> 
> 

@mikeki I actually wonder if we still need to have this fix in? Because this was believed to fix https://uniswapteam.slack.com/archives/C021SU4PMR7/p1711827069423259 at beginning, but the root cause turned out to be something else.

lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
```diff
@@ -168,7 +168,16 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
...
-          .filter((record) => !record.protocol || protocols.includes(record.protocol))
+          .filter(
+            (record) =>
+              !record.protocol ||
+              protocols.includes(record.protocol) ||
+              // we explicitly filter out the expired routes here,
+              // because "You can still update the expired items that are pending deletion,
+              // including changing or removing their TTL attributes"
+              // (https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html)
+              record.ttl >= Math.floor(Date.now() / 1000)
```
mikeki, line 179: I believe this should be:
```
record.ttl >= Math.floor(Date.now() / 1000) && (
  !record.protocol || protocols.includes(record.protocol)
)
```

jsy1218:
very wrong fix, now that we know why cached routes can become stale in RoutesDB

PR #545, opened 2024-04-02
[RPC Gateway] Add RPC Gateway specific Quote request count, fail count and latency metric

rollingtumbling:
We add more RPC gateway specific metrics that measures from the input side (GET QUOTE). Previously the metrics focused too much on the downstream side (call vendors).

Related dashboard is also updated.

This should be even more helpful to monitor our RPC gateway launches.

### DIFF
--- a/bin/stacks/rpc-gateway-dashboard.ts
+++ b/bin/stacks/rpc-gateway-dashboard.ts
@@ -196,6 +196,51 @@ function getCheckHealthMetricsForChain(chainId: ChainId) {
   return metrics
 }
 
+function getRpcGatewayQuoteCountForChain(chainId: ChainId) {
+  const metrics = []
+  metrics.push([
+    'Uniswap',
+    `RPC_GATEWAY_GET_QUOTE_REQUESTED_CHAINID: ${chainId}`,
+    'Service',
+    'RoutingAPI',
+    {
+      id: `quote_count_${chainId}`,
+      label: `Quote request count for ${ID_TO_NETWORK_NAME(chainId)}`
+    }
+  ])
+  return metrics
+}
+
+function getRpcGatewayQuoteLatencyForChain(chainId: ChainId) {
+  const metrics = []
+  metrics.push([
+    'Uniswap',
+    `RPC_GATEWAY_GET_QUOTE_LATENCY_CHAIN_${chainId}`,
+    'Service',
+    'RoutingAPI',
+    {
+      id: `quote_latency_${chainId}`,
+      label: `Quote latency for ${ID_TO_NETWORK_NAME(chainId)}`
+    }
+  ])
+  return metrics
+}
+
+function getRpcGatewayQuote5xxCountForChain(chainId: ChainId) {
+  const metrics = []
+  metrics.push([
+    'Uniswap',
+    `RPC_GATEWAY_GET_QUOTE_500_CHAINID: ${chainId}`,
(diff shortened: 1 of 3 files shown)


rollingtumbling:
* **#545** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/545?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

lib/handlers/quote/quote.ts
```diff
@@ -109,6 +110,9 @@ export class QuoteHandler extends APIGLambdaHandler<
           break
         case 500:
           metric.putMetric(`GET_QUOTE_500_CHAINID: ${chainId}`, 1, MetricLoggerUnit.Count)
+          if (useRpcGateway) {
```
jsy1218, line 113: Im actually confused why you need to add the `RPC_GATEWAY` metrics, meanwhile the existing `GET_QUOTE` metrics can cover both rpc gateway launch vs not launched yet cases.

bin/stacks/rpc-gateway-dashboard.ts
```diff
@@ -196,6 +196,51 @@ function getCheckHealthMetricsForChain(chainId: ChainId) {
   return metrics
 }
 
+function getRpcGatewayQuoteCountForChain(chainId: ChainId) {
+  const metrics = []
+  metrics.push([
+    'Uniswap',
+    `RPC_GATEWAY_GET_QUOTE_REQUESTED_CHAINID: ${chainId}`,
```
jsy1218, line 203: If you were to add the `RPC_GATEWAY` metrics and dashboard, you will need to add RPC gateway specific alarms as well.

jsy1218 (commented):
What Im trying to understand is that is there any anything missing from the current dashboard and alerts, that we need to add RPC gateway specific ones.

It seems like the prior incidents were nicely caught by the alerts right on time.

lib/handlers/quote/quote.ts
rollingtumbling: The reason is that I want to only log the amount of requests that are processed via RPC gateway code path. It's useful to monitor a chain that's partially launched with RPC gateway, where `GET_QUOTE_xxx` will be strictly larger than `RPC_GATEWAY_GET_QUOTE_xxx`.

bin/stacks/rpc-gateway-dashboard.ts
rollingtumbling: Do you mean adding alarm to the `RPC_GATEWAY_GET_QUOTE_500_CHAINID: xxx`? The total amount of quote requests going through RPC gateway doesn't need to trigger alarm for now.

jsy1218: Well, the alert should be resilient to the request volume. The alerts by default "treat missing data as missing", and therefore it's common to see alerts in the insufficient data state:



If we see the value in introducing the `RPC_GATEWAY_GET_QUOTE_500_CHAINID: xxx` metric, then I think it's also the case to introduce the corresponding alerts.

rollingtumbling: Thanks. I can make a follow up PR to add the alert.

jsy1218:
LGTM, let's not forget to add a follow-up PR on the RPC gateway alerts.

PR #548, opened 2024-04-03
[RPC Gateway] Reduce DB sync frequency and load

rollingtumbling:
It's very tricky to directly control the number of DB syncs.
The configs here only control the sync frequency within each lambda.
But if we have way too many lambdas, the number of sync can still be large.
However I still believe this adjustment is helpful when each lambda receives
larger number of requests.

We keep monitoring the DB sync load when we gradually increase the traffic
and launch to more chains. If it seems to go too high, we might need to
drop sync request with a probability.

Current DB read rate 

Current DB write rate


jsy1218 (commented):
@rollingtumbling and I have offline sync'ed. We think the current DB sync configs are probably ok for the low traffic chains. But for high traffic chains like Base, it might give some DB pressures. 

We think keep the current configs as default is good to progress through, meanwhile we might need to have a per-chain config, based on the chain traffic. This can be coded as needed, since we are launching per chain at a controlled percent, so we can extrapolate the DB sync traffic at 100% launch on that chain.

jsy1218 (approved):

PR #552, opened 2024-04-04
[RPC Gateway] Launch to 10% of Optimism

rollingtumbling:
[**RPC Gateway Deployment Overall Progress Track**](https://docs.google.com/spreadsheets/d/1_pbkfn_dXDLgChfdtkWZqxl2SLfKyYDLbXe-8z3YQmo/edit#gid=0)

Yesterday we deployed to 1% of Optimism. Since then

The success rate is unstable, but it's irrelevant to the RPC gateway (deployment time is show in the graph)

If we zoom out and look at latency change, there's no visible one in overall.

If we only look at RPC gateway specific request,

There's very few 5xx quote error. But if you look at the error from providers using RPC gateway path:

There are always random errors here and there in the last 24 hours. So I would consider this more of a downstream issue.

RPC gateway's quote p99 latency is about 2k to 4k ms

This aligns with overall quote p99 latency

PR #554, opened 2024-04-05
[RPC Gateway] Launch to 20% Optimism

rollingtumbling:
We deployed to 10% Optimism on yesterday 16:35 UTC-7 time with https://github.com/Uniswap/routing-api/pull/552. After the deployment we do see a bit of latency increase

It looks like both RPC gateway path and other code path has latency increases, because we even see the median latency has increased

Considering RPC gateway only counts for 10% of the traffic, if only RPC gateway code path have higher latency, the median latency shouldn’t change. That’s why I think it’s caused by something else.

Then I found that the latency for send() api from Infura, which is our Optimism’s major provider, did increase around the end of 18 o’clock. 

This should explain the QUOTE latency increase that happened about 2 hours after deployment.

However, just to be safe, we will try 20% launch and see if the latency continue to rise.

### DIFF
--- a/lib/config/rpcProviderProdConfig.json
+++ b/lib/config/rpcProviderProdConfig.json
@@ -19,7 +19,7 @@
   },
   {
     "chainId": 10,
-    "useMultiProviderProb": 1,
+    "useMultiProviderProb": 0.2,
     "providerInitialWeights": [1, 0, 0],
     "providerUrls": ["INFURA_10", "QUICKNODE_10", "ALCHEMY_10"]
   }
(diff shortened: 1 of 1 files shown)


rollingtumbling:
* **#554** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/554?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

jsy1218 (commented):
I'm not opposed to ramp up traffic in Optimism. But given that we already has sustained latency increase in Optimism, we'd probably want to slow it down a bit, just to double check the send latency increase is indeed irrelevant to the Opmitism launch.

I suggest we increase in 10% increment.

jsy1218 (approved):

PR #558, opened 2024-04-07
Finish dbSyncSampleProb

rollingtumbling:
Finish dbSyncSampleProb

Finish evaluateLatencySampleProb & healthCheckSampleProb


rollingtumbling:
* **#558** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/558?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -332,7 +336,7 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
       perf.latencyInMs = Date.now() - perf.startTimestampInMs
       this.checkLastCallPerformance(perf)
       this.updateHealthyStatus()
-      if (this.enableDbSync) {
```
jsy1218, line 335: You need to log two metrics to indicate the sampled DB Sync is indeed the desired sampling probability. See https://github.com/Uniswap/routing-api/blob/663523f6d7ed1e275e4b1fbec783279cb94246c6/lib/handlers/pools/provider-migration/v3/traffic-switch-v3-pool-provider.ts#L55 as an example.

jsy1218 (changes requested):
Is there a reason that rpcProviderProdConfig.json didn't explicitly configure those sampling probability? Without them, this code change is effectively no-op.

rollingtumbling:
It's indeed to be a no-op PR. The plan is to deploy to Sepolia first, then do another PR to only add the json config, then check if the reduction of DB sync event is as expected. But this PR cannot contain the config for Sepolia because that chains need to be supported first in https://github.com/Uniswap/routing-api/pull/561.

lib/rpc/SingleJsonRpcProvider.ts
rollingtumbling: Yes, now use two metrics as you suggested.

jsy1218 (approved):

PR #248, opened 2023-06-21
[ROUTE-30] Track infura RPC metrics

jsy1218:
## What?
From the last two Infura incidents on Polygon network outages, we realized we don't have nearly enough metrics on the downstream Infura. We are starting to track the Infura RPC metrics. This PR is the initial effort.

## Why?
We don't understand Infura call patterns. We don't know which Infura RPC methods have higher/lower success rate. We need to start gaining visibility before we can make the RPC provider fallback.

## How?
By inheriting/decorating `BaseProvider` with `InstrumentedEVMProvider` class.

## Testing?
Deployed to CDK via personal AWS account, and ran integ-test to see the new metrics pop up.

## Screenshots (optional)

Screenshot 2023-06-21 at 6 55 56 PM

## Anything Else?
Decided to not incorporate any potential fallback logic into this PR, since the fallback logic may depend on the current Infura success rate.

PR #559, opened 2024-04-08
[RPC Gateway] Add method name check before DB sync for non-shadow call

rollingtumbling:
We've seen cases where, for major provider, DB sync happens but `SingleJsonRpcProvider` doesn't have a latency record that belongs to the method name white list. Since we are only syncing latency for whitelisted methods with DB, this ends up being a wasted opportunity to sync. The result could be losing expected latency numbers in the DB for major providers.

We fix this issue by adding one more check on entering sync: The method name needs to be included in the whitelist.

Local test verified cases like mentioned above no longer happens.

Other small logging tweaks are added.

### DIFF
--- a/lib/rpc/ProviderStateSyncer.ts
+++ b/lib/rpc/ProviderStateSyncer.ts
@@ -38,7 +38,7 @@ export class ProviderStateSyncer {
     } catch (err: any) {
       this.log.error(`Failed to read from sync storage: ${JSON.stringify(err)}. Sync failed.`)
     }
-    this.log.debug({ storedState }, 'Loaded stored state')
+    this.log.debug({ storedState }, `${this.providerId}: Loaded stored state`)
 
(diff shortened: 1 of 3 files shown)


rollingtumbling:
* **#559** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/559?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

jsy1218 (approved):
imo, we should've started by optimistically overriding the RPC methods that have major traffic volume, instead of overriding every single RPC methods. Because of the prior work https://github.com/Uniswap/routing-api/pull/248, we already know the traffic volume of every single RPC across every chain.

rollingtumbling:
When we switch traffic or do fail over, we need to switch all of the RPC methods, right? So we have to intercept all RPC calls I guess?

PR #582, opened 2024-04-14
[RPC Gateway] Improve metrics logging

rollingtumbling:
Tiny adjustment on metrics reporting on RPC gateway provider call successes and fails.

PR #581, opened 2024-04-13
[RPC Gateway] Refactor the utils to get launched ChainId and provider names

rollingtumbling:

PR #583, opened 2024-04-14
[RPC Gateway Fallback Redesign] Implement the offline path

rollingtumbling:
Design doc: https://www.notion.so/uniswaplabs/Redesign-RPC-Gateway-Fallback-7c2ea784ae554f81a6d488bc52fab5ae?pvs=4

This PR implements the offline part, which includes
- Alarms
- Fallback-handling lambda
- Auto trigger of lambda from alarm
- Update DB health state

Tested in local env
- Using a test alarm (not the ones in this PR) to trigger an alarm
- Verified the DB entry was updated according to the state change of the alarm
- The correctness of the alarms definition in this PR is verified by creating manual graph using the expressions mentioned in this PR. That is, when alarm happens, DB state becomes `UNHEALTHY`, and when alarm recovers, DB state becomes `HEALTHY`. 

We specifically try to not touch the online serving part to minimize risk for causing an outage. As a result, we duplicated some of the code (mostly `ProviderHealthState.ts`, `ProviderHealthStateRepository.ts` and `ProviderHealthStateDynamoDbRepository.ts`) for the offline path. In the next PR, we will modify the online path, where these mentioned new code will be used (and the old code will be removed). For the same reason, a new DB table is used, which will replace the old DB table in the next PR when we modify the online path.

The merge of PR will not touch prod behavior. But to help us verify the behavior in prod, we use some very sensitive alarm configuration to trigger alarm frequently so the flow can be verified to be working in prod. We will change the alarm's sensitiveness in future PRs. (These alarms won't cause any ticket, BTW).


rollingtumbling:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/583?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#583** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/583?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#582** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/582?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#581** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/581?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

bin/stacks/routing-database-stack.ts
```diff
@@ -148,7 +153,7 @@ export class RoutingDatabaseStack extends cdk.NestedStack {
       }
     )
 
-    // Creates a Table for storing health state of RPC providers
+    // TODO(jie): Remove this table after we fully migrate to rpcProviderHealthStateDynamoDb
```
jsy1218, line 156: I would caution removing the table, because we've never explicitly set delete for retention policy on dynamo table:

Codewise, I mean we will have to specify destroy on removal, otherwise it's retain by default:
```diff
this.rpcProviderStateDynamoDb = new aws_dynamodb.Table(this, DynamoDBTableProps.RpcProviderStateDbTable.Name, {
      tableName: DynamoDBTableProps.RpcProviderStateDbTable.Name,
      partitionKey: {
        name: DynamoDBTableProps.RpcProviderStateDbTable.PartitionKeyName,
        type: AttributeType.STRING,
      },
      billingMode: BillingMode.PAY_PER_REQUEST,
      timeToLiveAttribute: DynamoDBTableProps.TTLAttributeName,
+     removalPolicy: RemovalPolicy.DESTROY,
    })
```

Since CDK relies on knowing the state of the DB, removing this DB might cause the next CDK deploy to fail in prod, and we will be stuck.

I recommend we just leave it as is, since we get the DDB bill pay per request.

bin/stacks/rpc-gateway-fallback-stack.ts
```diff
@@ -0,0 +1,147 @@
...
+    rpcHealthProviderStateDynamoDB.grantReadWriteData(lambdaRole)
+
+    const region = cdk.Stack.of(this).region
+
+    const providerFallbackLambda = new aws_lambda_nodejs.NodejsFunction(this, 'ProviderFallbackLambda', {
+      role: lambdaRole,
+      runtime: aws_lambda.Runtime.NODEJS_18_X,
+      entry: path.join(__dirname, '../../lib/rpc/handler/index.ts'),
+      handler: 'fallbackHandler',
+      timeout: cdk.Duration.seconds(5),
```
jsy1218, line 39: Is 5 seconds long enough? Since fallback stack is the offline path, the timeout can theoritically be longer than routing lambda (9 seconds) right?

bin/stacks/rpc-gateway-fallback-stack.ts
```diff
@@ -0,0 +1,147 @@
...
+      },
+      layers: [
+        aws_lambda.LayerVersion.fromLayerVersionArn(
+          this,
+          'InsightsLayer',
+          `arn:aws:lambda:${region}:580247275435:layer:LambdaInsightsExtension:14`
+        ),
+      ],
+      tracing: aws_lambda.Tracing.ACTIVE,
+      logRetention: RetentionDays.ONE_MONTH,
```
jsy1218, line 54: Do you actually need one month of log retention for the fallback lambda? It's a stateless lambda, in a sense that it persists the provider health into DDB, so in case something goes wrong with this lambda, it feels like we should be able to debug within a month.

lib/rpc/handler/FallbackHandler.ts
```diff
@@ -0,0 +1,81 @@
...
+  constructor(log: Logger) {
+    const dbTableName = process.env.PROVIDER_HEALTH_STATE_DB_TABLE_NAME
+    if (!dbTableName) {
+      throw new Error('Missing DB_TABLE_NAME in env var')
+    }
+    this.healthStateRepository = new ProviderHealthStateDynamoDbRepository(dbTableName, log)
+    this.log = log
+  }
+
+  get handler() {
```
jsy1218, line 30: It would be great if we are explicit in the return type, as well as the lambda event param type. Since this handler is the handler that inherits the lambda handler, being more strongly typed benefits from that.

jsy1218 (approved):
Overall LGTM, couple comments

cgkol (commented):
Looks good overall, i would also let Guillaume comment

bin/stacks/routing-database-stack.ts
rollingtumbling: OK I see what you mean. To be safe, I can leave the old table in the stack as it is and add comment saying it's obsolete.

bin/stacks/routing-database-stack.ts
```diff
@@ -158,5 +163,20 @@ export class RoutingDatabaseStack extends cdk.NestedStack {
...
       timeToLiveAttribute: DynamoDBTableProps.TTLAttributeName,
     })
+
+    // Creates a Table for storing health state of RPC providers
+    this.rpcProviderHealthStateDynamoDb = new aws_dynamodb.Table(
+      this,
+      DynamoDBTableProps.RpcProviderHealthStateDbTable.Name,
+      {
+        tableName: DynamoDBTableProps.RpcProviderHealthStateDbTable.Name,
+        partitionKey: {
```
uni-guillaume, line 173: no sort key ?

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
```diff
@@ -0,0 +1,44 @@
...
+import { ProviderHealthState } from './ProviderHealthState'
+
+export class ProviderHealthStateDynamoDbRepository implements ProviderHealthStateRepository {
+  private ddbClient: DynamoDB.DocumentClient
+
+  constructor(private dbTableName: string, private log: Logger) {
+    this.ddbClient = new DynamoDB.DocumentClient()
+  }
+
+  async read(providerId: string): Promise<ProviderHealthState | null> {
```
uni-guillaume, line 14: it looks like the provider health state is for all chain : should we handle the case where a provider is good at chain A but not for chain B?

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
```diff
@@ -0,0 +1,44 @@
...
+import Logger from 'bunyan'
+import { DynamoDB } from 'aws-sdk'
+import { DocumentClient } from 'aws-sdk/clients/dynamodb'
+import { ProviderHealthState } from './ProviderHealthState'
+
+export class ProviderHealthStateDynamoDbRepository implements ProviderHealthStateRepository {
+  private ddbClient: DynamoDB.DocumentClient
+
+  constructor(private dbTableName: string, private log: Logger) {
+    this.ddbClient = new DynamoDB.DocumentClient()
```
uni-guillaume, line 11: I would rather we inject a storage interface than newing up the DDB client here, it would increase the inversion of control and the unit testability

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
```diff
@@ -0,0 +1,44 @@
+import { ProviderHealthStateRepository } from './ProviderHealthStateRepository'
+import Logger from 'bunyan'
+import { DynamoDB } from 'aws-sdk'
+import { DocumentClient } from 'aws-sdk/clients/dynamodb'
+import { ProviderHealthState } from './ProviderHealthState'
+
+export class ProviderHealthStateDynamoDbRepository implements ProviderHealthStateRepository {
```
uni-guillaume, line 7: no unit test ?

lib/rpc/handler/FallbackHandler.ts
```diff
@@ -0,0 +1,97 @@
...
+
+      if (
+        (alarmEvent.previousState === 'OK' || alarmEvent.previousState === 'INSUFFICIENT_DATA') &&
+        alarmEvent.state === 'ALARM'
+      ) {
+        metric.putMetric(`RPC_GATEWAY_FALLBACK_${alarmEvent.providerId}_INTO_UNHEALTHY`, 1, MetricLoggerUnit.Count)
+        this.log.error(
+          `${alarmEvent.providerId} becomes UNHEALTHY due to ${alarmEvent.previousState}=>ALARM in ${alarmEvent.alarmName}`
+        )
+        await this.healthStateRepository.write(alarmEvent.providerId, ProviderHealthState.UNHEALTHY)
```
uni-guillaume, line 50: should this line come before the metric and log statement, in case it fails ?

lib/rpc/handler/FallbackHandler.ts
```diff
@@ -0,0 +1,97 @@
...
+      } else if (
+        alarmEvent.previousState === 'ALARM' &&
+        (alarmEvent.state === 'OK' || alarmEvent.state === 'INSUFFICIENT_DATA')
+      ) {
+        metric.putMetric(`RPC_GATEWAY_FALLBACK_${alarmEvent.providerId}_INTO_HEALTHY`, 1, MetricLoggerUnit.Count)
+        this.log.error(
+          `${alarmEvent.providerId} becomes HEALTHY due to ALARM=>${alarmEvent.state} in ${alarmEvent.alarmName}`
+        )
+        await this.healthStateRepository.write(alarmEvent.providerId, ProviderHealthState.HEALTHY)
```
uni-guillaume, line 59: same comment

lib/rpc/handler/FallbackHandler.ts
```diff
@@ -0,0 +1,97 @@
...
+    return metricScope((metricsLogger: MetricsLogger) => async (event: object): Promise<APIGatewayProxyResult> => {
+      metricsLogger.setNamespace('Uniswap')
+      metricsLogger.setDimensions({ Service: 'RoutingAPI' })
+      const metric = new AWSMetricsLogger(metricsLogger)
+
+      const alarmEvent = this.readAlarmEvent(event)
+      this.log.debug({ alarmEvent }, 'Parsed alarmEvent')
+
+      if (
+        (alarmEvent.previousState === 'OK' || alarmEvent.previousState === 'INSUFFICIENT_DATA') &&
```
uni-guillaume, line 43: not sure if the previous state really matters: what do we lose if we only write `if(alarmEvent.state === 'ALARM'` ?

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
rollingtumbling: There is `test/mocha/integ/rpc/ProviderHealthStateDynamoDbRepository.test.ts` down below.

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
rollingtumbling: No, every `Chain-Provider` pair has a health state.

The DB table looks like this:

lib/rpc/handler/FallbackHandler.ts
```diff
@@ -0,0 +1,97 @@
...
+        alarmEvent.state === 'ALARM'
+      ) {
+        metric.putMetric(`RPC_GATEWAY_FALLBACK_${alarmEvent.providerId}_INTO_UNHEALTHY`, 1, MetricLoggerUnit.Count)
+        this.log.error(
+          `${alarmEvent.providerId} becomes UNHEALTHY due to ${alarmEvent.previousState}=>ALARM in ${alarmEvent.alarmName}`
+        )
+        await this.healthStateRepository.write(alarmEvent.providerId, ProviderHealthState.UNHEALTHY)
+      } else if (
+        alarmEvent.previousState === 'ALARM' &&
+        (alarmEvent.state === 'OK' || alarmEvent.state === 'INSUFFICIENT_DATA')
```
uni-guillaume, line 53: same comment about the previous state. In this case, if the previous state was insufficient_data, then we would not turn back to green. (even though it the transition from ALARM -> INSUFFICIENT_DATA is captured by this `if` the transition from OK to INSUFFICIENT_DATA is not)

lib/rpc/handler/FallbackHandler.ts
```diff
@@ -0,0 +1,97 @@
...
+      const response = await handler(event)
+      this.log.debug({ response }, 'Response of fallback handler')
+      return response
+    }
+  }
+
+  readAlarmEvent(event: any): AlarmEvent {
+    // Example alarm name: "RoutingAPI-RpcGateway-ErrorRateAlarm-ChainId-56-Provider-QUIKNODE"
+    const alarmName: string = event.alarmData.alarmName
+    const alarmNameTokens = alarmName.split('-')
```
uni-guillaume, line 82: this works but it is a little brittle: is there any way we could use AWS tags instead ?

test/mocha/integ/rpc/ProviderHealthStateDynamoDbRepository.test.ts
```diff
@@ -0,0 +1,60 @@
...
+})
+
+describe('ProviderHealthStateDynamoDbRepository', () => {
+  setupTables(DB_TABLE)
+  const storage: ProviderHealthStateRepository = new ProviderHealthStateDynamoDbRepository(
+    DynamoDBTableProps.RpcProviderHealthStateDbTable.Name,
+    log
+  )
+
+  it('write state to DB then read from it', async () => {
```
uni-guillaume, line 44: this test does not test the functionality of an actual DDB, i suggest moving it to a unit test with a mocked DDB interface

test/mocha/unit/rpc/FallbackHandler.test.ts
```diff
@@ -0,0 +1,151 @@
...
+                "unit": "Count"
+              },
+              "returnData": false
+            }
+          ]
+        }
+      }
+    }`)
+
+  it('test readAlarmEvent', async () => {
```
uni-guillaume, line 118: the INSUFFICIENT DATA transitions are not tested.
You have
OK -> ALARM
ALARM -> OK

We also need
UD -> OK
UD -> ALARM
ALARM -> UD
OK -> UD

lib/rpc/handler/FallbackHandler.ts
rollingtumbling: I only want to take action when the state actually changes. This serves as a guard.

Also if you consider the "recover" case, we do need to consider previous state. For example, we don't want to update the DB state when states changes from "OK" to "INSUFFICIENT_DATA", but we do when states changes from "ALARM" to "INSUFFICIENT_DATA".

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
rollingtumbling: This whole class `ProviderHealthStateDynamoDbRepository` is exactly made for the purpose you mentioned. You will see in other PR (https://github.com/Uniswap/routing-api/pull/601) that this class is reused to access DB for the online path.

lib/rpc/handler/FallbackHandler.ts
rollingtumbling: Nice suggestion! Fixed.

lib/rpc/handler/FallbackHandler.ts
rollingtumbling: Fixed

test/mocha/integ/rpc/ProviderHealthStateDynamoDbRepository.test.ts
rollingtumbling: This test does use a real DynamoDB instance. It has to be run after starting a local DynamoDB instance. See `setupTables()` function where we configure DDB IP and port number.

lib/rpc/handler/FallbackHandler.ts
rollingtumbling: Unfortunately we have never used it in all of the metrics created from this repo. I can add a TODO here about it, but it's kind of out of the scope of this PR.

lib/rpc/handler/FallbackHandler.ts
rollingtumbling: It's a good question to discuss. There's no doubt about how we treat `OK` or `ALARM`, but for `INSUFFICIENT_DATA` it's up to some debate.

If we are being optimistic, we should treat `INSUFFICIENT_DATA` as benign. I think this aligns more accurately with our expected behaviors: When the services starts, it has no knowledge about provider's health state, but it will treat the provider as benign and use it anyway.

@cgkol what do you think about it? Can you share your comments?

lib/rpc/handler/FallbackHandler.ts
```diff
@@ -0,0 +1,97 @@
...
+export class FallbackHandler {
+  private healthStateRepository: ProviderHealthStateRepository
+  private log: Logger
+
+  constructor(log: Logger) {
+    const dbTableName = process.env.PROVIDER_HEALTH_STATE_DB_TABLE_NAME
+    if (!dbTableName) {
+      throw new Error('Missing DB_TABLE_NAME in env var')
+    }
+    this.healthStateRepository = new ProviderHealthStateDynamoDbRepository(dbTableName, log)
```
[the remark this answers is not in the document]
rollingtumbling: Currently in this repository there are no global DynamoDB client instance that we can use for injection. All usages of DynamoDB are using their own DDB client.

bin/stacks/routing-database-stack.ts
cgkol: i guess we can add timestamp to see latest changes right? or do we want to keep overriding?

lib/rpc/handler/FallbackHandler.ts
cgkol: I would rather that we change on insufficient_data, becuase I'd rather we dont get stuck on a bad state. However, given that when it starts it might take a while for it to get healthy, I think thats fair to just not change anything with `INSUFFICIENT_DATA`, just make sure its well document on the comments; assuming insufficient_data m

bin/stacks/routing-database-stack.ts
rollingtumbling: I don't see a downside for overriding. If we want to know what happened as history, we already published metrics for tracking all DB changes.

cgkol: we publish changes from tracking the metrics but not the DB changes right?

rollingtumbling: After a DB change, we will publish a metrics about that change so we can remember it in future (the metrics is specifically for logging DB change).

uni-guillaume: it depends if we ever considered queries that are chain wide or provider wide:
if you want to do `give me the status of all chain for a given provider`, then HK: provider, SK: chain
If you want to do `give the status of all providers for a given chain`, then HK: chain, SK: provider

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
uni-guillaume: that's integration test

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
uni-guillaume: not sure how that matters. The current way makes it impossible to unit test, you need an actual DDB

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
uni-guillaume: I see

lib/rpc/handler/FallbackHandler.ts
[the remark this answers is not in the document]
uni-guillaume: you can create the client where you need this class though: 
new FallbackHandler(log, new ProviderHealthStateDynamoDbRepository(new DDB Client()))

lib/rpc/handler/FallbackHandler.ts
cgkol: i think wheneve you ahve insufficient_data, default to previous state

lib/rpc/handler/FallbackHandler.ts
uni-guillaume: fair enough

test/mocha/integ/rpc/ProviderHealthStateDynamoDbRepository.test.ts
uni-guillaume: it's still an actual DDB, even if it is local. It makes the test unecessarily heavy IMHO.

bin/stacks/routing-database-stack.ts
rollingtumbling: Here all keys are a `Chain-Provider` pair so the keys are flat. The number of such pairs will be well below 50 so there's really no concern for the query efficiency.

In reality, one provider may have different infra for serving different chains, the outage of one chain from this provider doesn't necessarily mean its other chain will be impacted. So I think tracking health state at `Chain-Provider` pair level is an reasonable choice.

test/mocha/integ/rpc/ProviderHealthStateDynamoDbRepository.test.ts
rollingtumbling: Agreed, it's heavier than unit test...
Currently, using `dynamodb-local` is a universally accepted solution in `routing-api`, `api`, `data-api-graphql` and other repos. Fortunately we don't have big payload in the test so they run pretty fast so far.

test/mocha/unit/rpc/FallbackHandler.test.ts
rollingtumbling: UT for all these cases are added.

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
rollingtumbling: Although the folder name is `integ`, they are actually unit tests, just with the usage of DynamoDB.

lib/rpc/handler/FallbackHandler.ts
rollingtumbling: Thanks for the suggestion. The logic in `FallbackHandler.ts` are now fixed.

lib/rpc/handler/FallbackHandler.ts
rollingtumbling: Addressed in other comments. Code logic is fixed as suggested.

lib/rpc/handler/FallbackHandler.ts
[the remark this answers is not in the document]
rollingtumbling: I agree inverse of control is a better concept. Currently it's OK to have multiple DynamoDB client embedded in multiple classes. Regarding of testing, it's easy enough to mock a class member (such as a DynamoDB client) in typescript unit test, so a re-rejection isn't very necessary. I believe for some other languages, using injection might solve more pain points for writing UT.

cgkol (approved):

uni-guillaume (approved):

lib/rpc/handler/FallbackHandler.ts
```diff
@@ -88,4 +87,79 @@ export class FallbackHandler {
...
+      ProviderHealthiness.UNHEALTHY
+    )
+  }
+
+  private async updateDbItemWhenOk(alarmEvent: AlarmEvent): Promise<HealthinessUpdate> {
+    const state: ProviderHealthState | null = await this.healthStateRepository.read(alarmEvent.providerId)
+    if (state === null || state.healthiness === ProviderHealthiness.HEALTHY) {
+      return new HealthinessUpdate(ProviderHealthiness.HEALTHY, ProviderHealthiness.HEALTHY)
+    }
+    if (!state.ongoingAlarms.includes(alarmEvent.alarmName)) {
```
cgkol, line 150: how is this even possible btw?

lib/rpc/handler/FallbackHandler.ts
```diff
@@ -0,0 +1,165 @@
...
+export interface AlarmEvent {
+  alarmName: string
+  state: string
+  previousState: string // Only for logging purpose. Not used in any logic.
+  providerId: string
+  reason: string // Only for logging purpose. Not used in any logic
+}
+
+export class HealthinessUpdate {
+  constructor(public oldHealthiness: ProviderHealthiness, public newHealthiness: ProviderHealthiness) {}
```
cgkol, line 21: Why does the constructor need to have both passed in? My undersatnding is that the old change should be part of the constructor and the new state should be part of isChanged?

lib/rpc/handler/FallbackHandler.ts
```diff
@@ -0,0 +1,165 @@
...
+      return new HealthinessUpdate(ProviderHealthiness.UNHEALTHY, ProviderHealthiness.UNHEALTHY)
+    }
+    const newOngoingAlarms = state.ongoingAlarms.filter((alarmName) => alarmName !== alarmEvent.alarmName)
+    const newState: ProviderHealthState = {
+      healthiness: newOngoingAlarms.length === 0 ? ProviderHealthiness.HEALTHY : ProviderHealthiness.UNHEALTHY,
+      ongoingAlarms: newOngoingAlarms,
+      version: state.version + 1,
+    }
+    await this.healthStateRepository.update(alarmEvent.providerId, newState)
+    return new HealthinessUpdate(
```
cgkol, line 160: strange here that we pass two status, we should follow the comment above around how the class is constrcucted. Also, not sure if you even need a class? You only use it once and you can compare at the time you are calling "isChanged" anyway with a simple comparator

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
```diff
@@ -0,0 +1,86 @@
...
+      ConditionExpression: CONDITION_EXPRESSION,
+    }
+    await this.ddbClient.update(updateParams).promise()
+  }
+
+  private getExpressionAttributeValues(state: ProviderHealthState): { [key: string]: any } {
+    let attributes: { [key: string]: any } = {}
+    attributes[':healthiness'] = state.healthiness
+    attributes[':ongoingAlarms'] = state.ongoingAlarms
+    attributes[':baseVersion'] = state.version - 1
```
cgkol, line 82: why do we need to the base version and nee version? also if the differene is only one, do we ever need it stored?

test/mocha/unit/rpc/FallbackHandler.test.ts
```diff
@@ -220,41 +80,220 @@ describe('FallbackHandler', () => {
...
+        healthiness: ProviderHealthiness.UNHEALTHY,
+        ongoingAlarms: ['alarm2'],
+        version: 2,
+      })
+      expect(healthinessUpdate).deep.equals(
+        new HealthinessUpdate(ProviderHealthiness.UNHEALTHY, ProviderHealthiness.UNHEALTHY)
+      )
+    })
+
+    it('has previous alarm, new alarm repeats, all ongoing alarm cleared', async () => {
```
cgkol, line 270: if it repeats as an error state, it should stay on right? are you just assuming it can only flip/flop?

lib/rpc/handler/FallbackHandler.ts
```diff
@@ -88,4 +87,79 @@ export class FallbackHandler {
...
+        metric.putMetric(`RPC_GATEWAY_FALLBACK_${alarmEvent.providerId}_INTO_HEALTHY`, 1, MetricLoggerUnit.Count)
+        this.log.error(
+          `${alarmEvent.providerId} becomes HEALTHY due to ${alarmEvent.previousState}=>OK in ${alarmEvent.alarmName}`
+        )
+      }
+    }
+    return healthinessUpdate
+  }
+
+  private async updateDbItemWhenAlarm(alarmEvent: AlarmEvent): Promise<HealthinessUpdate> {
```
cgkol, line 114: it is strange for these two functions to be separate, instead of just using `alarmEvent.state` and doing the selection with the two functions

cgkol (commented):
Few comments

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
rollingtumbling: I should've added more comments in the code. It's used for dealing with write conflicts, that is, two fallback lambdas trying to write to the same DB primary key at the same time.

All entry updates are designed to increase the version number by 1, so we expect the existing DB version number to be minus 1 of the version number in the update. However, if it's not the case, it means a write conflict has occurred and the latter update should cancel and retry with another lambda call.

I've verified that, if the lambda throws exception, AWS will retry another 2 times with the same input (alarm event object), with each having a delay of about 60 seconds. So in the retry, it should be able to finish the write.

test/mocha/unit/rpc/FallbackHandler.test.ts
rollingtumbling: This UT is for testing `updateDbItemWhenOk()` function, which means we received a `xxx` => `OK` event. The input event's `state` is `OK` as shown in line 210. It's not "new alarm repeats", it should be read as "new alarm event repeats". I've fixed the description.

lib/rpc/handler/FallbackHandler.ts
rollingtumbling: Yes the logic you mentioned is actually in `processAlarm()`. The reason for such is to easy unit testing because the `AWSMetricsLogger` is somewhat not very easy to mock in UT if you are not using the global `metric` object, in my case. So I want to add another layer to not let it bother me in writing UT.

cgkol: Changing logic for testing is not optimal; You can just create a function thats called updateItemWhenEvent and pass that in right?

lib/rpc/handler/FallbackHandler.ts
cgkol: I expalined myself above we can keep it there

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
rollingtumbling: If the conditional expression `'#version = :baseVersion'` evaluates to false, `this.ddbClient.update()` will throw an exception

```
ConditionalCheckFailedException: The conditional request failed
```
There's no `try...catch` to catch this error so it will pop up as a lambda invocation error.

Optimistic locking is a better choice here because the probability of having concurrent update is rare in our use case. The total amount of alarm state changes regarding a chain-provider pair is small and will probably not even causing the spawn of another lambda (however, if it happens and we do nothing about, some provider might never recover). In comparison, adding a lock (pessimistic locking) is a much less efficient way. It's only be worth it if we want absolute transactional updates or the frequency of conflict is high...both are not the case here.

lib/rpc/handler/FallbackHandler.ts
rollingtumbling: Refactored this part of code. The readability should be much improved by now.

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
cgkol: i still dont think its a good solution, but dont want to block you

cgkol (approved):

lib/rpc/ProviderHealthStateDynamoDbRepository.ts
rollingtumbling: Can you share why you don't think it's a good solution?

PR #592, opened 2024-04-17
chore: also consider the header for the request source and app version

mikeki:
# Consider the request source and app version

Include a consideration for the request source and app version in the protocol selection


lib/handlers/quote/quote.ts
```diff
@@ -232,24 +232,28 @@ export class QuoteHandler extends APIGLambdaHandler<
     }
 
     let protocols: Protocol[] = []
+
+    const isMobileRequest = ['uniswap-ios', 'uniswap-android'].includes(params.event.headers['x-request-source'] ?? params.requestQueryParams.source ?? '');
+    // const appVersion = params.event.headers['x-app-version'];
```
cgkol, line 237: we probably want to change this and add some smol unit tests right?

lib/handlers/quote/quote.ts
```diff
@@ -232,25 +233,25 @@ export class QuoteHandler extends APIGLambdaHandler<
     }
 
     let protocols: Protocol[] = []
+
+    const isMobileRequest = ['uniswap-ios', 'uniswap-android'].includes(params.event.headers['x-request-source'] ?? params.requestQueryParams.source ?? '');
+    const appVersion = params.event.headers['x-app-version'];
+    // We will exclude V2 if isMobile and the appVersion is not present or is lower or equal than 1.24
+    const excludeV2 = isMobileRequest && (appVersion === undefined || semver.lte(appVersion, '1.24'))
```
jsy1218, line 240: I thought app version 1.24 is a good version, and should not be excluded. This might be important, because over half of install base was on 1.24 just a week ago, so worth triple checking.

test/jest/unit/handlers/quote.test.ts
```diff
@@ -0,0 +1,156 @@
...
+                requestSource,
+                appVersion,
+                undefined
+              )
+            ).toEqual([Protocol.V2, Protocol.V3, Protocol.MIXED]);
+          });
+        });
+      });
+    });
+  });
```
mikeki, line 155: @adjkant are these the kind of versions that I can expect being sent in the x-app-version header from the mobile app?

adjkant: Yes except for .test!

mikeki: would you ever have something like `.build1`?

jsy1218 (approved):

cgkol (approved):
✅ thank you for the tests!

lib/handlers/quote/quote.ts
```diff
@@ -642,6 +624,50 @@ export class QuoteHandler extends APIGLambdaHandler<
...
+              protocols.push(Protocol.MIXED)
+            }
+            break
+          default:
+            return []
+        }
+      }
+
+      return protocols
+    } else if (!forceCrossProtocol) {
```
dannydaniil, line 664: @mikeki why is this line v3 only?

mikeki: thats a good question, I'm actually not sure, it is probably because in the past we didnt have v2 support on all chains and we were default to v3 only. This is how it was before though, so I didnt change the business logic there

PR #593, opened 2024-04-17
fix: current quoter out-of-gas quote comparison

jsy1218:
Once SOR returns per-quote-call gasUsed and gasLimit (https://github.com/Uniswap/smart-order-router/pull/534), then routing-api can calculate per-quote call failure is due to likely out-of-gas error. 

I was deciding between using gasUsed and gasLimit vs the encoded error from the return data, but the [simulation](https://www.tdly.co/shared/simulation/39de04bb-d794-43b0-aed2-8ea31957ca77) shows me that the encoded error from the return data might not be accurate. For this simulation, the error indicates [SLP](https://docs.uniswap.org/contracts/v3/reference/error-codes), which is not true:

 ```
cast 4byte-decode 0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000000353504c0000000000000000000000000000000000000000000000000000000000
1) "Error(string)"
"SPL"
```

Now testing coverage added.

### DIFF
--- a/lib/handlers/quote/provider-migration/v3/traffic-switch-on-chain-quote-provider.ts
+++ b/lib/handlers/quote/provider-migration/v3/traffic-switch-on-chain-quote-provider.ts
@@ -4,7 +4,7 @@ import { ChainId, Currency, CurrencyAmount } from '@uniswap/sdk-core'
 import { ProviderConfig } from '@uniswap/smart-order-router/build/main/providers/provider'
 import { QUOTE_PROVIDER_TRAFFIC_SWITCH_CONFIGURATION } from '../../util/quote-provider-traffic-switch-configuration'
 import { BigNumber } from 'ethers'
-import { NEW_QUOTER_DEPLOY_BLOCK } from '../../../../util/onChainQuoteProviderConfigs'
+import { LIKELY_OUT_OF_GAS_THRESHOLD, NEW_QUOTER_DEPLOY_BLOCK } from '../../../../util/onChainQuoteProviderConfigs'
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: lib/handlers/quote/provider-migration/v3/traffic-switch-on-chain-quote-provider.ts [hunk 1]
```diff
@@ -287,12 +287,24 @@ export class TrafficSwitchOnChainQuoteProvider implements IOnChainQuoteProvider
...
 
         if (!(currentQuote.quote ?? BigNumber.from(0)).eq(targetQuote.quote ?? BigNumber.from(0))) {
+          // This is to deal with the case, where some quotes from the quoter multicalls failed,
+          // but still above the success rate threshold (https://github.com/Uniswap/smart-order-router/blob/main/src/providers/on-chain-quote-provider.ts#L496)
+          // so that the current quote failed, but the view-only quoter is more gas-efficient, and can return the quote.
+          if (!currentQuote.quote && targetQuote.quote) {
+            const gasEstimate = currentQuote.gasEstimate?.toNumber() ?? 0
+            const gasLimit = currentQuote.gasLimit?.toNumber() ?? 0
+
+            if (gasLimit - gasEstimate <= LIKELY_OUT_OF_GAS_THRESHOLD[this.chainId]) {
```
rollingtumbling, line 297: What if `this.chainId` is not included in the `LIKELY_OUT_OF_GAS_THRESHOLD` ?

rollingtumbling (approved):

PR #600, opened 2024-04-18
fix: when protocols is undefined throw error, otherwise query empty array

mikeki:
ForceCrossProtocol flag would keep an empty array of protocols, and it would query alpha router using that empty array. SOR will then determine the protocol that it must request.

I accidentally tarted throwing 400s on this case since I wrongly assumed that an empty array of protocols was a failure case


rollingtumbling (approved):
Thanks for the fix.
Please fix lint.

PR #601, opened 2024-04-18
[RPC Gateway Fallback Resign] Implement the online path

rollingtumbling:
Following the design https://www.notion.so/uniswaplabs/Redesign-RPC-Gateway-Fallback-7c2ea784ae554f81a6d488bc52fab5ae?pvs=4 and previous PR for offline path https://github.com/Uniswap/routing-api/pull/583, we modify the online path to support auto fallback and recover based on DB states.

A few highlights in this PR
- Previous "health score" based calculation is totally removed. Online path will only publish metrics and listen to DB for provider health state updates
- The auto fallback and recover is able to turn on/off at a per chain basis. This can help us to test it fully, for example, when deployed for a testnet with prod traffic
- All UTs have been updated. All unused util classes has been removed

Tested on local env:
- Verified the offline path to be working: Auto update DB state based on alarm state
- Verified the online path to be working: See a switch of provider selection after the preferred provider's DB state becomes unhealthy. Also see the reverse switch of provider selection when then provider's DB state resumes to healthy
  - This is observed using the new updated dashboards which shows DB state change and `SingleJsonProvider`'s state change history, as well as the number of selection of providers when serving RPC calls.

An example of provider switch, showing switching from Nirvana to Infura when serving RPC requests after Nirvana becomes unhealthy.

Another example where we let the primary provider fail, then we see the traffic goes to the backup provider, then we let the primary provider recover, then the traffic goes back to the primary provider:


rollingtumbling:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/601?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#601** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/601?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#583** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/583?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#594](https://github.com/Uniswap/routing-api/pull/594) <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/594?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#582** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/582?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#581** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/581?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

bin/stacks/routing-database-stack.ts
```diff
@@ -153,17 +148,6 @@ export class RoutingDatabaseStack extends cdk.NestedStack {
       }
     )
 
-    // TODO(jie): Remove this table after we fully migrate to rpcProviderHealthStateDynamoDb
```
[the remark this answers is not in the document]
rollingtumbling: Thanks for the suggestion. I'd better keep the table then. I will just leave it there and not refer it from other parts of the code.

rollingtumbling:
You are not looking at the latest version of the PR.
`-1` means the provider can be used as a fallback. 

 *  Positive value represents provider weight when selecting from different healthy providers.
   *  0 means this provider will never be able to be selected.
   *  -1 means this provider will only be able to be selected when no healthy provider has positive weights. In that case, the first healthy provider with -1 will be selected.
   *  Not providing this argument means using -1 for all weight values.

jsy1218 (approved):
Overall LGTM

lib/config/rpcProviderProdConfig.json
```diff
@@ -37,9 +37,10 @@
   },
   {
     "chainId": 42161,
-    "useMultiProviderProb": 0.1,
-    "providerInitialWeights": [1, 0, 0, 0],
-    "providerUrls": ["INFURA_42161", "QUICKNODE_42161", "NIRVANA_42161", "ALCHEMY_42161"]
+    "useMultiProviderProb": 1,
+    "providerInitialWeights": [1, -1, -1, -1],
+    "providerUrls": ["NIRVANA_42161", "ALCHEMY_42161", "INFURA_42161", "QUICKNODE_42161"],
```
cgkol, line 42: you are about to send 100% of traffic to nirvana for 42161?

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -339,80 +269,43 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
...
+            this.syncAndUpdateProviderHealthState()
           }
         }
       }
-      this.lastCallTimestampInMs = perf.startTimestampInMs
     }
   }
 
-  private async syncAndUpdateProviderState() {
+  private async syncAndUpdateProviderHealthState() {
```
cgkol, line 286: so every `dbSyncSampleProb` you check to see if the health state has changed? and if so, update it? Do we still need the boolean `this.syncingDb`? Whats the expected logic here?

cgkol (commented):
Will look at the tests after this review

also curious, how did you get nirvana to start failing to test?

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -339,80 +269,43 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
...
-      if (newState !== null) {
-        // Update health state
-        this.healthScoreAtLastSync = newState.healthScore
-        this.healthScore = this.healthScoreAtLastSync
-        this.log.debug(`${this.providerId}: Synced with storage: new health score ${this.healthScore}`)
-        this.updateHealthyStatus()
-
-        // Update latency stat
-        this.updateLatencyStat(newState)
+      const healthStateFromDb = await this.healthStateRepository.read(this.providerId)
```
uni-guillaume, line 288: note for future improvement: each chain could have several alarm (P50 latency, P90 latency, success rate, volume...) mapped to columns in the DB. From this we could compute a score that gives you the weight directly here

lib/config/rpcProviderProdConfig.json
rollingtumbling: Sorry this is for debug. Reverted already.

lib/rpc/SingleJsonRpcProvider.ts
rollingtumbling: Both are needed and they serve different purposes.

`this.syncingDb` serves as a lock to ensure that only 1 DB read will happen even if the lambda is processing multiple RPC requests. Without it, before the read of first DB read finishes, more DB read will be made. 

`dbSyncSampleProb` is to reduce the spike of huge number of lambda doing DB read at the same time. The use of `this.syncingDb` and its sync interval still cannot prevent all lambdas doing DB read at the same time, which could be a dangerous load to DB.

This is essentially the idea I discussed in the design doc https://www.notion.so/uniswaplabs/Rate-Limit-DB-Sync-in-RPC-Gateway-40fec8f3ab7e49fca8b3471586b55970?pvs=4

rollingtumbling:
> Will look at the tests after this review
> 
> also curious, how did you get nirvana to start failing to test?

There are two options
1. Directly modify the DB entry
2. Using a lower threshold for latency alarm, then call the endpoint of my local environment to make RPC calls to Nirvana. Usually the latency will be high enough to trigger the alarm, then the DB is updated as well.

lib/rpc/SingleJsonRpcProvider.ts
rollingtumbling: Thanks for the suggestion. Currently the DB schema looks like this:

```json
{
  "chainIdProviderName": {
    "S": "11155111_ALCHEMY"
  },
  "healthiness": {
    "S": "UNHEALTHY"
  },
  "ongoingAlarms": {
    "L": [
      {
        "S": "RoutingAPI-RpcGateway-LatencyAlarm-ChainId-11155111-Provider-ALCHEMY"
      }
    ]
  },
  "version": {
    "N": "29"
  }
}
```

So all ongoing alarms have been put inside the DB entry.

PR #613, opened 2024-04-22
fix: improve quote provider logging

jsy1218:
We want to be able to different the quoter protocol version by explicitly adding the `Mixed` vs `V3` metrics wording. This is done primary through SOR https://github.com/Uniswap/smart-order-router/pull/547


cgkol (approved):
are we seeing errors?

jsy1218:
> are we seeing errors?
> 
> 

Errors rate are the same in prod last week, but I realized when we shadow quote, our metrics don't differentiate between v3 quoter and mixed quoter. Since we are only replacing v3 quoter, we should separate metrics to see the effects only shadow v3 quoter.

PR #622, opened 2024-04-22
fix: activity id to chain id quote injector and lambda handler event

jsy1218:
During container injector, we instantiate a new global logger with [injector-sor](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/injector-sor.ts#L132-L137), without request id, or any identifier that can identifier which request it's part of:

```
    const log: Logger = bunyan.createLogger({
      name: this.injectorName,
      serializers: bunyan.stdSerializers,
      level: bunyan.INFO,
    })
    setGlobalLogger(log)
```

The reason for this is because, [buildContainerInjected()](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/injector-sor.ts#L131C16-L131C38) is called as part of  [QuoteHandlerInjector('quoteInjector').build()](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/index.ts#L13), but we only start to get the request id from the lambda event within [new QuoteHandler('quote', quoteInjectorPromise)
](https://github.com/Uniswap/routing-api/blob/main/lib/handlers/index.ts#L14C18-L15C1), within [injector](https://github.com/Uniswap/routing-api/blob/e9d62b4bc4b2886fd9eccbe4cc6f94a4a494dfe8/lib/handlers/quote/injector.ts#L32):

```
  public async getRequestInjected(
    containerInjected: ContainerInjected,
    _requestBody: void,
    requestQueryParams: QuoteQueryParams,
    _event: APIGatewayProxyEvent,
    context: Context,
    log: Logger,
    metricsLogger: MetricsLogger
  ): Promise<RequestInjected<IRouter<AlphaRouterConfig | LegacyRoutingConfig>>> {
    const requestId = context.awsRequestId
  ...

log = log.child({
      serializers: bunyan.stdSerializers,
      level: logLevel,
      requestId,
      quoteId,
      tokenInAddress,
      chainId: tokenInChainId,
      tokenOutAddress,
      amount,
      type,
      algorithm,
      gasToken,
    })
    setGlobalLogger(log)
}
```

So this means we need to manually generate a new random identifier within injector-sor `activityId`, and pass into injector, so that we have the way to identify which request is this container injection from. In addition, because not every request needs the container injection, the activity id has to be optional. I verified in my local and can now see activityId popping up and tieing up the loggings within the same request:

Screenshot 2024-04-22 at 4 55 27 PM
Screenshot 2024-04-22 at 4 55 35 PM

I also tested non-cold start scenario, and can see there's no activity id, only request id:

### DIFF
--- a/lib/handlers/injector-sor.ts
+++ b/lib/handlers/injector-sor.ts
@@ -65,6 +65,7 @@ import {
   RETRY_OPTIONS,
   SUCCESS_RATE_FAILURE_OVERRIDES,
 } from '../util/onChainQuoteProviderConfigs'
+import { v4 } from 'uuid/index'
 
 export const SUPPORTED_CHAINS: ChainId[] = [
   ChainId.MAINNET,
@@ -120,6 +121,7 @@ export interface ContainerInjected {
   dependencies: {
     [chainId in ChainId]?: ContainerDependencies
   }
+  activityId?: string
 }
 
 export abstract class InjectorSOR<Router, QueryParams> extends Injector<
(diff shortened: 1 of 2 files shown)


mikeki (approved):

jsy1218:
### Merge activity

* **Apr 22, 8:06 PM EDT**: @jsy1218 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/622).

PR #623, opened 2024-04-23
chore: log pool parsing latency metrics

jsy1218:
I want to make sure that the pool parsing in memory is not taking long, so logging that. And for each s3 object download and parsing, I also log the latency as part of the logging info, so that when we debug by activity id, we can see each s3 object download latency.


mikeki (approved):

jsy1218:
### Merge activity

* **Apr 23, 12:30 PM EDT**: @jsy1218 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/623).

PR #651, opened 2024-05-07
feat: create new pool bucket with compressed files

jsy1218:
We noticed the lambda cold start is due to the s3 large pool file download. We have decided to compress the pool files during the pool cron jobs, and download the compressed files.

We are now uploading compressed pool files into a new pool s3 bucket, with the new object key prefix `poolCacheGzip.json`. Then during the same cron job instance, we will also decompress the gzipped file, and check sum the original subgraph pools vs the decompressed pools to ensure there's no data loss at all.

In addition, we measure the following important metrics:
1. compression ratio between original file and compressed file. So far the compression ratio looks around 4:1:

2. checksum metrics to ensure every compressed pool has lossless data:

3. latency for just decompressing the pool file object in-memory. This is important as in a subsequent PR, meanwhile the s3 download will be faster in terms of the network round-trip, it will be somewhat offset by the in-memory decompression. So far it looks around 100-200ms, which is a good tradeoff, for average of 3s s3 download latency:


bin/stacks/routing-caching-stack.ts
```diff
@@ -64,7 +67,22 @@ export class RoutingCachingStack extends cdk.NestedStack {
       expiration: cdk.Duration.days(365 * 10),
     })
 
+    this.poolCacheBucket3.addLifecycleRule({
+      enabled: true,
+      // This isn't the right fix in the long run, but it will prevent the outage that we experienced when the V2 pool
```
mikeki, line 72: Could we remove this comment and say see comment above or something like that?

lib/cron/cache-pools.ts
```diff
@@ -55,23 +56,124 @@ const handler: ScheduledHandler = metricScope((metrics) => async (event: EventBr
...
-    .promise()
+  const beforeDecompress = Date.now()
+  const decompressedPools = zlib.inflateSync(compressedPools)
+  metric.putMetric(`${metricPrefix}.decompression_latency`, Date.now() - beforeDecompress)
+
+  const parsedPools: V2SubgraphPool[] | V3SubgraphPool[] = JSON.parse(decompressedPools.toString())
+
+  checkSum(pools, parsedPools, log, metric, chainId, protocol)
+
+  const [result, newResult] = await Promise.all([
```
mikeki, line 81: Can we do this before the check?

mikeki (approved):

PR #662, opened 2024-05-08
feat: tune Celo batch param for gas limit and multichunk size

jsy1218:
We are ready to tune the gas batch params in Celo to see if it can speed up. If we check the gas used on P97, P98, P99 and max levels:

we see the gas used drops significantly at P98 level to around ~3M per call. Since current batch param sets gas limit per call at 5M, we can drop it to around ~3M. Then we can increase the multichunk size to make sure each multicall can shoehorn more quote calls, reducing the number of RPC calls to multicall. We need to ensure that gas limit per call * multichunk size remains constant, or smaller, but not larger. So we choose 31200000 * 16 = 499200000, or 4.9M.

We choose Celo first, since it's lowest risk, meanwhile the traffic volume should be good enough to see quote endpoint latency improvements, if any.


jsy1218:
* **#662** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/662?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mikeki (approved):

PR #667, opened 2024-05-09
fix: no on-chain quote retry for celo

jsy1218:
I think there's no need to retry failed on-chain quoter for Celo. If we look at max:

this tells me in the worst case scenario, we always exhausted all the attempts (2), and also the first time batch param vs the low success rate param do not improve success rate at all (0% -> 0%).

Then looking at p95 and p98:

It tells me, sometimes there's retry due to low success rate, otherwise it doesn't retry and just give up. In both cases, the success rate is low at 0%.

Eventually look at min:

This tells me that the min quote success rate still stays at 0%, despite no retry loop.

All those metrics together tells me, whether we retry or not, for those failed on-chain quoting, they will fail regardless of retry attempts. Therefore, there's no point in retrying, and hopefully with no retry, this brings celo quote endpoint latency down.

### DIFF
--- a/lib/util/onChainQuoteProviderConfigs.ts
+++ b/lib/util/onChainQuoteProviderConfigs.ts
@@ -32,7 +32,7 @@ export const RETRY_OPTIONS: { [chainId: number]: AsyncRetry.Options | undefined 
     maxTimeout: 1000,
   },
   [ChainId.CELO]: {
-    retries: 2,
+    retries: 0,
     minTimeout: 100,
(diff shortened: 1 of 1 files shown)


jsy1218:
* **#667** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/667?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mikeki (approved):

PR #674, opened 2024-05-13
feat: tune OP batch param for gas limit and multichunk size

jsy1218:
We are ready to tune the gas batch params in Op to see if it can speed up. If we check the gas used on P50:

optimistic P50 gas limit is around ~80k per call, meanwhile non-optimistic P50 is around 150k per call.  Right now we have 1.2M gas limit per call, which is too high. We decrease the gas limit respectively, so that we can increase the multichunk size for less multicall RPC calls over the network.

And if we look at the retry loop at P90:

less than 10% of the calls are retried even once, so we should go for some aggressive but still reasonable gas tuning on OP.


jsy1218:
* **#674** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/674?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mikeki (approved):

PR #683, opened 2024-05-14
fix: increase base gas limit to 100k

jsy1218:
Base on the optimistic cached routes is lowered to 75k gas limit per call, and we see increase in retries:

Since non-optimistic path of 200k gas limit doesn't cause increase in retries, so somewhere between 75k-200k is good.

At P75 gas, I see the non-optimistic gas limit is ~100k:

So Im thinking to increase optimistic gas limit from 75k to 100k as a heuristic.

Overall, by lowering to 75k gas limit doesnt change the onchain quoting latency, so it means even though we have more retries, the increased batch size offset the retries in the overall latencies:

which means we should either expect to see improved latency from this PR, or still no latency improvement on Base.


jsy1218:
* **#683** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/683?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

cgkol (approved):

PR #687, opened 2024-05-15
fix: measure rpc methods latencies right around the call

jsy1218:
In order to get the most accurate latency measure for each RPC method, we will shuffle the latency measurement to be exactly where the RPC method invocation happens.

In addition, we will log high latency alchemy RPC method request payload, so as to collaborate to debug.

### DIFF
--- a/lib/rpc/SingleJsonRpcProvider.ts
+++ b/lib/rpc/SingleJsonRpcProvider.ts
@@ -256,9 +256,19 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
       startTimestampInMs: Date.now(),
     }
     try {
+      // TODO: consolidate start time here and above
       perf.startTimestampInMs = Date.now()
       const result = await fn(...args)
       perf.latencyInMs = Date.now() - perf.startTimestampInMs
+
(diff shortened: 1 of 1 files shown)


lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -256,7 +256,20 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
       startTimestampInMs: Date.now(),
     }
     try {
-      return await fn(...args)
+      // TODO: consolidate start time here and above
+      perf.startTimestampInMs = Date.now()
+      const result = await fn(...args)
+      perf.latencyInMs = Date.now() - perf.startTimestampInMs
```
mikeki, line 262: if the call fails this could be 0, right? since we are not longer updating the latencyInMs before logging?

jsy1218: I think in that case, we won't log the latency metrics, because the code checks whether the perf object is successful or not https://github.com/Uniswap/routing-api/blob/main/lib/rpc/SingleJsonRpcProvider.ts#L163.

But admittedly, I think the implementation is a bit over-engineered, so I may choose to refactor/simplify later.

mikeki (approved):

PR #736, opened 2024-06-13
feat: Introduce `GraphQLTokenFeeFetcher` (no traffic)

xrsv:
Introducing a `GraphQLTokenFeeFetcher` (implementation + tests only for now, no live traffic)
This fetcher will use GraphQL endpoint to fetch token fee data, and will fallback to OnChain fetching if something goes wrong or to fetch any remaining tokens that couldn't be fetched from graphQL.

Did setup env variables `UNI_GRAPHQL_ENDPOINT` / `UNI_GRAPHQL_HEADER_ORIGIN` for GraphQL in routing-api

A separate PR will follow for shadow traffic,

#### Testing
- added unit/integration tests
- ran e2e tests with 100% GraphQL traffic

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,6 +10,9 @@ jobs:
   test:
     name: Run tests
     runs-on: ubuntu-latest
+    env:
+      UNI_GRAPHQL_ENDPOINT: ${{ secrets.UNI_GRAPHQL_ENDPOINT }}
+      UNI_GRAPHQL_HEADER_ORIGIN: ${{ secrets.UNI_GRAPHQL_HEADER_ORIGIN }}
 
     steps:
       - name: Checkout Repo
--- a/lib/graphql/graphql-client.ts
+++ b/lib/graphql/graphql-client.ts
@@ -1,0 +1,47 @@
+import axios, { AxiosRequestConfig, AxiosResponse } from 'axios';
+
+import { GraphQLResponse } from './graphql-schemas';
+
+/* Interface for accessing any GraphQL API */
+export interface IGraphQLClient {
+  fetchData<T>(query: string, variables?: { [key: string]: any }): Promise<T>;
+}
+
+/* Implementation of the IGraphQLClient interface to give access to any GraphQL API */
+export class GraphQLClient implements IGraphQLClient {
+  private readonly endpoint: string;
+  private readonly headers: { [key: string]: string };
+
+  constructor(endpoint: string, headers: { [key: string]: string }) {
+    this.endpoint = endpoint;
+    this.headers = headers;
+  }
+
+  async fetchData<T>(query: string, variables: { [key: string]: any } = {}): Promise<T> {
+    const requestConfig: AxiosRequestConfig = {
+      method: 'POST',
+      url: this.endpoint,
+      headers: this.headers,
+      data: { query, variables },
+    };
+
+    let response: AxiosResponse<GraphQLResponse<T>>;
+    try {
+      response = await axios.request(requestConfig);
+    } catch (error) {
+      if (axios.isAxiosError(error)) {
+        throw new Error(`HTTP error! status: ${error.response?.status}`);
+      } else {
+        throw new Error(`Unexpected error: ${error}`);
+      }
+    }
+
+    const responseBody = response.data;
+    if (responseBody.errors) {
+      throw new Error(`GraphQL error! ${JSON.stringify(responseBody.errors)}`);
+    }
+
+    return responseBody.data;
+  }
+}
+
--- a/lib/graphql/graphql-client.ts
+++ b/lib/graphql/graphql-client.ts
@@ -1,20 +1,20 @@
-import axios, { AxiosRequestConfig, AxiosResponse } from 'axios';
+import axios, { AxiosRequestConfig, AxiosResponse } from 'axios'
 
-import { GraphQLResponse } from './graphql-schemas';
+import { GraphQLResponse } from './graphql-schemas'
 
 /* Interface for accessing any GraphQL API */
 export interface IGraphQLClient {
-  fetchData<T>(query: string, variables?: { [key: string]: any }): Promise<T>;
+  fetchData<T>(query: string, variables?: { [key: string]: any }): Promise<T>
 }
 
 /* Implementation of the IGraphQLClient interface to give access to any GraphQL API */
 export class GraphQLClient implements IGraphQLClient {
-  private readonly endpoint: string;
-  private readonly headers: { [key: string]: string };
+  private readonly endpoint: string
+  private readonly headers: { [key: string]: string }
 
   constructor(endpoint: string, headers: { [key: string]: string }) {
-    this.endpoint = endpoint;
-    this.headers = headers;
+    this.endpoint = endpoint
+    this.headers = headers
   }
 
   async fetchData<T>(query: string, variables: { [key: string]: any } = {}): Promise<T> {
@@ -23,25 +23,24 @@ export class GraphQLClient implements IGraphQLClient {
       url: this.endpoint,
       headers: this.headers,
       data: { query, variables },
-    };
+    }
 
-    let response: AxiosResponse<GraphQLResponse<T>>;
+    let response: AxiosResponse<GraphQLResponse<T>>
     try {
-      response = await axios.request(requestConfig);
+      response = await axios.request(requestConfig)
     } catch (error) {
       if (axios.isAxiosError(error)) {
-        throw new Error(`HTTP error! status: ${error.response?.status}`);
+        throw new Error(`HTTP error! status: ${error.response?.status}`)
       } else {
-        throw new Error(`Unexpected error: ${error}`);
+        throw new Error(`Unexpected error: ${error}`)
       }
     }
 
-    const responseBody = response.data;
+    const responseBody = response.data
     if (responseBody.errors) {
-      throw new Error(`GraphQL error! ${JSON.stringify(responseBody.errors)}`);
+      throw new Error(`GraphQL error! ${JSON.stringify(responseBody.errors)}`)
     }
 
-    return responseBody.data;
+    return responseBody.data
   }
 }
-
--- a/lib/graphql/graphql-client.ts
+++ b/lib/graphql/graphql-client.ts
@@ -9,11 +9,7 @@ export interface IGraphQLClient {
 
 /* Implementation of the IGraphQLClient interface to give access to any GraphQL API */
 export class GraphQLClient implements IGraphQLClient {
-
-  constructor(
-    private readonly endpoint: string,
-    private readonly headers: Record<string, string>
-  ) {}
+  constructor(private readonly endpoint: string, private readonly headers: Record<string, string>) {}
 
   async fetchData<T>(query: string, variables: { [key: string]: any } = {}): Promise<T> {
     const requestConfig: AxiosRequestConfig = {
--- a/lib/graphql/graphql-client.ts
+++ b/lib/graphql/graphql-client.ts
@@ -19,9 +19,14 @@ export class GraphQLClient implements IGraphQLClient {
       data: { query, variables },
     }
 
-    let response: AxiosResponse<GraphQLResponse<T>>
     try {
-      response = await axios.request(requestConfig)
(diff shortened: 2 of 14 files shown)


lib/handlers/injector-sor.ts
```diff
@@ -233,17 +234,28 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
...
             new NodeJSCache(new NodeCache({ stdTTL: 30000, useClones: false }))
           )
-          const tokenPropertiesProvider = new TokenPropertiesProvider(
-            chainId,
-            new NodeJSCache(new NodeCache({ stdTTL: 30000, useClones: false })),
-            tokenFeeFetcher
-          )
+          let tokenPropertiesProvider: TokenPropertiesProvider
+          // Gradually switch to graphql token fee fetcher - Start with 10%.
+          if (Math.random() <= 0.1) {
```
xrsv, line 244: will update this with something similar to [traffic-switch-on-chain-quote-provider.ts](https://github.com/Uniswap/routing-api/blob/267ec26f90eee67a0610a1bbd6181587b450105d/lib/handlers/quote/provider-migration/v3/traffic-switch-on-chain-quote-provider.ts)

lib/graphql/graphql-client.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/* Implementation of the IGraphQLClient interface to give access to any GraphQL API */
+export class GraphQLClient implements IGraphQLClient {
+  private readonly endpoint: string
+  private readonly headers: { [key: string]: string }
+
+  constructor(endpoint: string, headers: { [key: string]: string }) {
+    this.endpoint = endpoint
+    this.headers = headers
+  }
```
mikeki, line 18: style suggestion, we can simplify this with the syntactic sugar:

```
constructor(
  private readonly endpoint: string,
  private readonly headers: Record<string, string>
) {}
```

I'll leave this decision up to you though, whatever seems more readable in your opinion

lib/graphql/graphql-client.ts
```diff
@@ -0,0 +1,46 @@
+import axios, { AxiosRequestConfig, AxiosResponse } from 'axios'
+
+import { GraphQLResponse } from './graphql-schemas'
+
+/* Interface for accessing any GraphQL API */
+export interface IGraphQLClient {
+  fetchData<T>(query: string, variables?: { [key: string]: any }): Promise<T>
+}
```
mikeki, line 8: In SOR we use this library for graphql requests: https://github.com/jasonkuhrt/graphql-request

I really have no strong opinion either way, but I just wanted to throw it out there too in case you were not aware :)

lib/graphql/graphql-token-fee-fetcher.ts
```diff
@@ -0,0 +1,75 @@
...
+export class GraphQLTokenFeeFetcher implements ITokenFeeFetcher {
+  private readonly graphQLProvider: IUniGraphQLProvider
+  private readonly onChainFeeFetcherFallback: ITokenFeeFetcher
+  private readonly chainId: ChainId
+
+  constructor(chainId: ChainId, rpcProvider: BaseProvider) {
+    this.graphQLProvider = new UniGraphQLProvider()
+    this.onChainFeeFetcherFallback = new OnChainTokenFeeFetcher(chainId, rpcProvider)
+    this.chainId = chainId
+  }
```
mikeki, line 31: I'd suggest we pass the OnChainTokenFeeFetcher as a constructor argument, rather than initialized here, probably the same with the graphql provider, to follow Inversion of Control patterns

lib/graphql/graphql-provider.ts
```diff
@@ -0,0 +1,61 @@
...
+  // - more details: https://github.com/Uniswap/data-api-graphql/blob/main/graphql/schema.graphql
+}
+
+/* Implementation of the UniGraphQLProvider interface to give access to Uniswap GraphQL API */
+export class UniGraphQLProvider implements IUniGraphQLProvider {
+  private readonly endpoint = process.env.UNI_GRAPHQL_ENDPOINT!
+  private readonly headers = {
+    Origin: process.env.UNI_GRAPHQL_HEADER_ORIGIN!,
+    'Content-Type': 'application/json',
+  }
```
mikeki, line 26: We probably need to add these env variables to github actions for the tests to pass, but we should also have landed on the fallback for those tests that are failing?

mikeki (commented):
It's looking good overall, some minor comments :)

jsy1218:
Pasting my [comments](https://github.com/Uniswap/smart-order-router/pull/603#discussion_r1638471926) from old PR. I think we want to shadow sample GQL traffic first, with FOT fee comparison. We aren't in rush to switch 10% (or whatever) traffic to GQL

xrsv:
> Pasting my [comments](https://github.com/Uniswap/smart-order-router/pull/603#discussion_r1638471926) from old PR. I think we want to shadow sample GQL traffic first, with FOT fee comparison. We aren't in rush to switch 10% (or whatever) traffic to GQL

makes sense - I'll remove the traffic switch part from this PR and will add that logic to a different PR

lib/graphql/graphql-provider.ts
xrsv: initialization is outside our try/catch which makes sense, we want to fail here and catch this in tests.
will add envs to gh actions

lib/graphql/graphql-client.ts
xrsv: thanks - wasn't aware. Will keep as is for now as it's a simple implementation + tested.
might refactor in the future though if we need any extra functionality

lib/handlers/injector-sor.ts
xrsv: sampling will happen in a follow up PR

lib/graphql/graphql-provider.ts
xrsv: added - please tal that looks ok
tested against persoanl aws as well

.github/workflows/test.yml
```diff
@@ -10,6 +10,9 @@ jobs:
   test:
     name: Run tests
     runs-on: ubuntu-latest
+    env:
+      UNI_GRAPHQL_ENDPOINT: ${{ secrets.UNI_GRAPHQL_ENDPOINT }}
+      UNI_GRAPHQL_HEADER_ORIGIN: ${{ secrets.UNI_GRAPHQL_HEADER_ORIGIN }}
```
jsy1218, line 15: nit: let's also add those two env vars into the [README](https://github.com/Uniswap/routing-api/blob/main/README.md)

lib/graphql/graphql-client.ts
jsy1218: I think routing-api is better off just using the axios library for remote calls, unless we see a strong benefit of DevEx in using a particular library like typechain for ease of RPC calls compositions.

otoh, if in future, we have many GQL upstream dependencies, we can push the data team to implement a custom GQL client for every upstream to consume, including us. Until then, I'm in favor of less external library dependencies in routing-api/SOR, especially we already have many library dependencies.

lib/graphql/graphql-client.ts
```diff
@@ -0,0 +1,40 @@
...
+        throw new Error(`Unexpected error: ${error}`)
+      }
+    }
+
+    const responseBody = response.data
+    if (responseBody.errors) {
+      throw new Error(`GraphQL error! ${JSON.stringify(responseBody.errors)}`)
+    }
+
+    return responseBody.data
```
jsy1218, line 38: nit: I think you can put response variable declaration inside the try block, so that you don't have to re-assign.

```
try {
      const response: AxiosResponse<GraphQLResponse<T>> = await axios.request(requestConfig);
    const responseBody = response.data
    if (responseBody.errors) {
      throw new Error(`GraphQL error! ${JSON.stringify(responseBody.errors)}`)
    }
    return responseBody.data
} catch (error) {
      if (axios.isAxiosError(error)) {
        throw new Error(`HTTP error! status: ${error.response?.status}`)
      } else {
        throw new Error(`Unexpected error: ${error}`)
      }
}
```

lib/graphql/graphql-client.ts
```diff
@@ -0,0 +1,40 @@
...
+      url: this.endpoint,
+      headers: this.headers,
+      data: { query, variables },
+    }
+
+    let response: AxiosResponse<GraphQLResponse<T>>
+    try {
+      response = await axios.request(requestConfig)
+    } catch (error) {
+      if (axios.isAxiosError(error)) {
```
jsy1218, line 26: nit: we probably want to add both GQL call success and failure metrics, so that we can measure the GQL call success rate from routing-api perspective.

lib/graphql/graphql-provider.ts
```diff
@@ -0,0 +1,61 @@
...
+  private client: IGraphQLClient
+
+  constructor() {
+    this.client = new GraphQLClient(this.endpoint, this.headers)
+  }
+
+  /* Convert ChainId to a string recognized by data-graph-api graphql endpoint.
+   *  GraphQL Chain Enum located here: https://github.com/Uniswap/data-api-graphql/blob/main/graphql/schema.graphql#L155
+   *  */
+  private _chainIdToGraphQLChainName(chainId: ChainId): string | undefined {
```
jsy1218, line 36: I understand that GQL only supports mainnet for FOT fee fetching, but I think from routing-api perspective, we probably don't have to code this logic ourselves, because:
1. it's easy to forget that we also have a per-chain gatekeeping
2. on the top-level token fee fetcher that we plan to implement, we can also another in-memory cache layer, so that the first time we hit GQL and get error due to chain not supported, we can fallback to SOR and cache the result for the entire duration of the lambda cycle. in other words, as long as we implement the top-level token fee fetcher right, we won't DDOS against GQL endpoint on the unsupported chains.

lib/graphql/graphql-provider.ts
```diff
@@ -0,0 +1,61 @@
...
+    const query = GRAPHQL_QUERY_TOKEN_INFO_BY_ADDRESS_CHAIN
+    const variables = { chain: this._chainIdToGraphQLChainName(chainId), address: address }
+    return this.client.fetchData<TokenInfoResponse>(query, variables)
+  }
+
+  async getTokensInfo(chainId: ChainId, addresses: string[]): Promise<TokensInfoResponse> {
+    const query = GRAPHQL_QUERY_MULTIPLE_TOKEN_INFO_BY_CONTRACTS
+    const contracts = addresses.map((address) => ({
+      chain: this._chainIdToGraphQLChainName(chainId),
+      address: address,
```
jsy1218, line 56: We probably will need to limit the number of addresses to get from GQL.

lib/graphql/graphql-client.ts
xrsv: we already log success/failure metrics in `graphql-token-fee-fetcher.ts` level, do you think we need extra logging here?

lib/graphql/graphql-provider.ts
xrsv: no matter the number of addresses, it's a single call to GQL. Not sure if we need a limit here, at this point at least

lib/graphql/graphql-client.ts
jsy1218: sorry didn't see, nah we don't need.

lib/graphql/graphql-provider.ts
jsy1218: let's test it out during the next PR, and see if we need a limit.

lib/graphql/graphql-provider.ts
xrsv: for # 2, this situation would be handled by our onchain fallback.
for # 1, I agree but this method here is necessary as we want to translate [sdk-core/ChainId enum](https://github.com/Uniswap/sdk-core/blob/main/src/chains.ts#L1) to a string that matches [data-api-graphql/Chain enum](https://github.com/Uniswap/data-api-graphql/blob/main/graphql/schema.graphql#L155). Unfortunately diff enums are used in different places in our repos. We should probably ask data team to provide (and maintain) this conversion method for us eventually.

mikeki (approved):
Overall LGTM

jsy1218 (approved):

jsy1218:
### Merge activity

* **Jun 17, 12:32 PM EDT**: @jsy1218 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/736).

PR #745, opened 2024-06-14
fix: less max splits and higher distribution percent for less routes zksync

jsy1218:


jsy1218:
* **#745** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/745?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

mikeki (approved):

zhongeric (approved):

lib/handlers/shared.ts
```diff
@@ -93,8 +93,8 @@ export const DEFAULT_ROUTING_CONFIG_BY_CHAIN = (chainId: ChainId): AlphaRouterCo
         },
         maxSwapsPerPath: 3,
         minSplits: 1,
-        maxSplits: 7,
-        distributionPercent: 5,
+        maxSplits: 2,
+        distributionPercent: 25,
```
cgkol, line 97: this pr now affects all chains (mainnet too). I think we want to case it for zksync only right? cc: @jsy1218 @zhongeric 

I was wondering why our latency wnet down too.. but this explains it

xrsv: i was also checking the latency metrics today, and ended up to this PR as well.
Agree this needs to apply to zksync only

jsy1218: fixed in https://github.com/Uniswap/routing-api/pull/746/files

PR #750, opened 2024-06-17
feat: Start GQL fee fetcher shadow sampling at 0.5% of traffic

xrsv:
Starting shadow sampling for `GraphQLTokenFeeFetcher` at `0.5%` of traffic.
This is only enabled on mainnet and, with recent caching changes, `0.5%` of traffic should be around `20 RPM` of sampling traffic (`avg 4k requests / min * 0.005 = 20 RPM`) (stats pulled from [here](https://us-east-2.console.aws.amazon.com/cloudwatch/home?region=us-east-2#metricsV2?graph=~(metrics~(~(~'Uniswap~'TokenFeeFetcherFetchFeesSuccess~'Service~'RoutingAPI)~(~'.~'TokenFeeFetcherFetchFeesFailure~'.~'.)~(~'.~'TokenPropertiesProviderBatchGetCacheHit~'.~'.))~view~'timeSeries~stacked~false~region~'us-east-2~stat~'SampleCount~period~60)&query=~'*7bUniswap*2cService*7d*20TokenPropertiesProviderBatchGetCacheHit))

Also adding a generic `TrafficSwitcher<T>` to handle traffic switching/sampling on 2 implementations of any type `T`.

### DIFF
--- a/lib/handlers/injector-sor.ts
+++ b/lib/handlers/injector-sor.ts
@@ -71,6 +71,9 @@ import { v4 } from 'uuid/index'
 import { chainProtocols } from '../cron/cache-config'
 import { Protocol } from '@uniswap/router-sdk'
 import { UniJsonRpcProvider } from '../rpc/UniJsonRpcProvider'
+import { GraphQLTokenFeeFetcher } from '../graphql/graphql-token-fee-fetcher'
+import { UniGraphQLProvider } from '../graphql/graphql-provider'
+import { TrafficSwitcherITokenFeeFetcher } from '../util/traffic-switch/traffic-switcher-i-token-fee-fetcher'
 
 export const SUPPORTED_CHAINS: ChainId[] = [
   ChainId.MAINNET,
@@ -231,7 +234,22 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
             sourceOfTruthPoolProvider: noCacheV3PoolProvider,
           })
 
-          const tokenFeeFetcher = new OnChainTokenFeeFetcher(chainId, provider)
+          const onChainTokenFeeFetcher = new OnChainTokenFeeFetcher(chainId, provider)
+          const graphQLTokenFeeFetcher = new GraphQLTokenFeeFetcher(
+            new UniGraphQLProvider(),
+            onChainTokenFeeFetcher,
+            chainId
+          )
+          const trafficSwitcherTokenFetcher = new TrafficSwitcherITokenFeeFetcher(
+            'TokenFetcherExperiment',
+            onChainTokenFeeFetcher,
+            graphQLTokenFeeFetcher,
+            'onChainTokenFeeFetcher',
+            'graphQLTokenFeeFetcher',
+            0.0,
+            0.05
+          )
+
           const tokenValidatorProvider = new TokenValidatorProvider(
             chainId,
             multicall2Provider,
@@ -240,7 +258,7 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
           const tokenPropertiesProvider = new TokenPropertiesProvider(
             chainId,
             new NodeJSCache(new NodeCache({ stdTTL: 30000, useClones: false })),
-            tokenFeeFetcher
+            trafficSwitcherTokenFetcher
           )
           const underlyingV2PoolProvider = new V2PoolProvider(chainId, multicall2Provider, tokenPropertiesProvider)
           const v2PoolProvider = new CachingV2PoolProvider(
--- a/lib/handlers/injector-sor.ts
+++ b/lib/handlers/injector-sor.ts
@@ -240,15 +240,16 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
             onChainTokenFeeFetcher,
             chainId
           )
-          const trafficSwitcherTokenFetcher = new TrafficSwitcherITokenFeeFetcher(
-            'TokenFetcherExperiment',
-            onChainTokenFeeFetcher,
-            graphQLTokenFeeFetcher,
-            'onChainTokenFeeFetcher',
-            'graphQLTokenFeeFetcher',
-            0.0,
-            0.005
-          )
+          const trafficSwitcherTokenFetcher = new TrafficSwitcherITokenFeeFetcher('TokenFetcherExperiment', {
+            control: onChainTokenFeeFetcher,
+            treatment: graphQLTokenFeeFetcher,
+            aliasControl: 'onChainTokenFeeFetcher',
+            aliasTreatment: 'graphQLTokenFeeFetcher',
+            customization: {
+              pctEnabled: 0.0,
+              pctShadowSampling: 0.005,
+            },
+          })
 
           const tokenValidatorProvider = new TokenValidatorProvider(
             chainId,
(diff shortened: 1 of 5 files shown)


lib/util/traffic-switch/traffic-switcher.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  constructor(
+    experimentName: string,
+    implementationA: T,
+    implementationB: T,
+    aliasA: string,
+    aliasB: string,
+    pctTrafficSwitchToB: number = 0.0,
+    pctSamplingToCompare: number = 0.0
+  ) {
```
mikeki, line 33: Here's a very neat Typescript style that I've seen in many repos and which I think you might like, It's a pattern that the AWS CDK and SDK use, as well as SST, and I think once you get into the groove of it, it's super helpful.

```
export interface TrafficSwitcherProps<TExperiment> {
  control: TExperiment;
  treatment: TExperiment;
  aliasControl: string,
  aliasTreatment: string,
  trafficSwitch: TrafficSwitchProps;
}

export interface TrafficSwitchProps {
  pctEnabled: number;
  pctShadowSampling: number;
}
```

then in this constructor:
```
  constructor(experimentName: string, props: TrafficSwitcherProps) {
    // here you map your props to your local variables.
   }
```

lib/util/traffic-switch/traffic-switcher-i-token-fee-fetcher.ts
```diff
@@ -0,0 +1,72 @@
...
+    experimentName: string,
+    implementationA: ITokenFeeFetcher,
+    implementationB: ITokenFeeFetcher,
+    aliasA: string,
+    aliasB: string,
+    pctTrafficSwitchToB: number,
+    pctSamplingToCompare: number
+  ) {
+    super(experimentName, implementationA, implementationB, aliasA, aliasB, pctTrafficSwitchToB, pctSamplingToCompare)
+  }
```
mikeki, line 18: This constructor can be ignored, since you are not passing any new information, the constructors are inherited as well.

lib/handlers/injector-sor.ts
```diff
@@ -231,7 +234,22 @@ export abstract class InjectorSOR<Router, QueryParams> extends Injector<
...
+          )
+          const trafficSwitcherTokenFetcher = new TrafficSwitcherITokenFeeFetcher(
+            'TokenFetcherExperiment',
+            onChainTokenFeeFetcher,
+            graphQLTokenFeeFetcher,
+            'onChainTokenFeeFetcher',
+            'graphQLTokenFeeFetcher',
+            0.0,
+            0.005
+          )
```
mikeki, line 251: finally here we just need to update how we initialize this class:

```
          const trafficSwitcherTokenFetcher = new TrafficSwitcherITokenFeeFetcher(
            'TokenFetcherExperiment',
            {
               control: onChainTokenFeeFetcher,
               treatment: graphQLTokenFeeFetcher,
               aliasControl: 'onChainTokenFeeFetcher',
               aliasTreatment: 'graphQLTokenFeeFetcher',
               {
                 pctEnabled: 0.0,
                 pctShadowSampling: 0.005,
               },
          })
```

test/mocha/unit/traffic-switch/traffic-switcher-i-token-fee-fetcher.test.ts
```diff
@@ -0,0 +1,292 @@
...
+import { OnChainTokenFeeFetcher, TokenFeeMap } from '@uniswap/smart-order-router/build/main/providers/token-fee-fetcher'
+import { BigNumber } from 'ethers'
+import { expect } from 'chai'
+import { MetricLoggerUnit } from '@uniswap/smart-order-router'
+import { TrafficSwitcher } from '../../../../lib/util/traffic-switch/traffic-switcher'
+
+describe('TrafficSwitcherITokenFeeFetcher', () => {
+  let spy: SinonSpy
+
+  const methodFetchFeesReturnFee100 = async (addresses: string[], _?: ProviderConfig): Promise<TokenFeeMap> => {
```
mikeki, line 15: could we add some tests to the base class as well? :p

mikeki (approved):

lib/util/traffic-switch/traffic-switcher.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+  }
+
+  protected isSamplingEnabled(): boolean {
+    return this.pctSamplingToCompare > 0
+  }
+
+  protected shouldSample(): boolean {
+    return this.isSamplingEnabled() && Math.random() < this.pctSamplingToCompare
+  }
```
mikeki, line 108: This will end up sampling on every method call rather than throughout the life of the request. This is okay for this use-case, but in a real experiment we will want to do this sampling in a more deterministic way, perhaps by hashing the requestId?

lib/util/traffic-switch/traffic-switcher.ts
```diff
@@ -0,0 +1,139 @@
...
+          [methodA(), this.aliasA, methodName],
+          [methodB(), this.aliasB, methodName],
+        ])
+
+        // Compare the results if a comparison method is provided
+        if (results && results.length == 2) {
+          try {
+            if (comparisonMethod) {
+              comparisonMethod(results[0], results[1])
+            }
```
mikeki, line 76: `comparisonMethod && comparisongMethod(results[0], results[1])` could also work if you like that

lib/util/traffic-switch/traffic-switcher.ts
xrsv: good call - currently logic is to sample/traffic switch at the per call/method level (`trafficSwitchMethod`) 
I agree probably in the future we'll need something stickier and requestId is a good candidate.
I will add a note + TODO for this, as it might require a bit more thinking to get it right.

xrsv: I see `TrafficSwitcher` as a work in progress class that will get more complete over time (e.g. your idea, plus we could add timing comparison etc).

test/mocha/unit/traffic-switch/traffic-switcher-i-token-fee-fetcher.test.ts
xrsv: base class is abstract, so this is the actual testing it mocking the TrafficSwitcherITokenFeeFetcher implementation.
Happy to add more tests here though

xrsv:
@mikeki that was good feedback - thanks
tried to address everything and looks much better

test/mocha/unit/traffic-switch/traffic-switcher-i-token-fee-fetcher.test.ts
mikeki: yeah, I was thinking of just creating an implementation TestTrafficSwitcher to test the methods of the base class in isolation, under the assumption that the `TokenFeeFetcher` traffic switch will be removed at the end of the experiment, when that happens then the base class will not have any tests.

lib/util/traffic-switch/traffic-switcher.ts
mikeki: yup yup, I agree, just doing some brainstorming here about the directions we could take this, but leaving them as good to haves for now

mikeki (approved):

PR #752, opened 2024-06-18
fix: graphql endpoint might return null feeData

xrsv:
From sampling traffic identified a small number of cases where graphql returns `feeData: null` and causing exception.
This was handled eventually with onChain fallback, but we want to handle correctly and keep sampling/comparing control and treatment.

### DIFF
--- a/lib/graphql/graphql-token-fee-fetcher.ts
+++ b/lib/graphql/graphql-token-fee-fetcher.ts
@@ -36,10 +36,12 @@ export class GraphQLTokenFeeFetcher implements ITokenFeeFetcher {
     try {
       const tokenFeeResponse: TokensInfoResponse = await this.graphQLProvider.getTokensInfo(this.chainId, addresses)
       tokenFeeResponse.tokens.forEach((token) => {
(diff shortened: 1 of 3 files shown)


lib/graphql/graphql-token-fee-fetcher.ts
```diff
@@ -36,10 +36,12 @@ export class GraphQLTokenFeeFetcher implements ITokenFeeFetcher {
     try {
       const tokenFeeResponse: TokensInfoResponse = await this.graphQLProvider.getTokensInfo(this.chainId, addresses)
       tokenFeeResponse.tokens.forEach((token) => {
-        if (token.feeData.buyFeeBps || token.feeData.sellFeeBps) {
+        if (token.feeData && (token.feeData.buyFeeBps || token.feeData.sellFeeBps)) {
```
mikeki, line 39: completely a nit, but this could also be: 
`if (token.feeData?.buyFeeBps || token.feeData?.sellFeeBps) {`

mikeki (approved):

PR #753, opened 2024-06-18
fix: update token fetcher sampling comparison logic to ignore empty/0 fees

xrsv:
Update on the sampling comparison logic between OnchainFeeFetcher and GraphQLFeeFetcher to ignore entries that are empty or have 0 fees and pollute comparison logs.

example
`TrafficSwitcherITokenFeeFetcher compareResultsForFetchFees: Different fee for address 0x4da27a545c0c5b758a6ba100e3a049001de870f5:  in control: {"buyFeeBps":{"type":"BigNumber","hex":"0x00"},"sellFeeBps":{"type":"BigNumber","hex":"0x00"}} and treatment: {}`

### DIFF
--- a/lib/util/traffic-switch/traffic-switcher-i-token-fee-fetcher.ts
+++ b/lib/util/traffic-switch/traffic-switcher-i-token-fee-fetcher.ts
@@ -31,19 +31,28 @@ export class TrafficSwitcherITokenFeeFetcher extends TrafficSwitcher<ITokenFeeFe
       return
     }
 
-    // We have results from both implementations, compare them as a whole
-    const identical = JSON.stringify(resultA) === JSON.stringify(resultB)
+    // We have results from both implementations, compare them as a whole.
+    // Before comparison, do some cleaning and keep only entries with a fee != 0.
+    // This is needed as different implementations can return empty/null entry, or entry with 0 fees.
+    const cleanResult = (result: TokenFeeMap): TokenFeeMap =>
+      Object.entries(result)
+        .filter(([_, v]) => !v.buyFeeBps?.eq(0) || !v.sellFeeBps?.eq(0))
+        .reduce((acc, [k, v]) => ({ ...acc, [k]: v }), {})
+    const cleanedResultA = cleanResult(resultA)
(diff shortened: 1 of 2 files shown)


lib/util/traffic-switch/traffic-switcher-i-token-fee-fetcher.ts
```diff
@@ -31,19 +31,28 @@ export class TrafficSwitcherITokenFeeFetcher extends TrafficSwitcher<ITokenFeeFe
       return
     }
 
-    // We have results from both implementations, compare them as a whole
-    const identical = JSON.stringify(resultA) === JSON.stringify(resultB)
+    // We have results from both implementations, compare them as a whole.
+    // Before comparison, do some cleaning and keep only entries with a fee != 0.
+    // This is needed as different implementations can return empty/null entry, or entry with 0 fees.
+    const cleanResult = (result: TokenFeeMap): TokenFeeMap =>
```
jsy1218, line 37: this feels slightly over complicated. if we can just filter out the fee === undefined and fee === 0, then both results are sanitized to compare

test/mocha/unit/traffic-switch/traffic-switcher-i-token-fee-fetcher.test.ts
```diff
@@ -294,4 +309,50 @@ describe('TrafficSwitcherITokenFeeFetcher', () => {
       MetricLoggerUnit.Count
     )
   })
+
+  it('sampling traffic 100% should get results from both (Identical empty/0 fee), and return Current impl result', async () => {
```
jsy1218, line 313: it seems like the `cleanResult` is well tested. Im not going to make it a blocking comment, if that;s the case.

jsy1218 (approved):

lib/util/traffic-switch/traffic-switcher-i-token-fee-fetcher.ts
xrsv: we can discuss this - i wanted to only make changes in the comparison method first, then we can figure out how to clean results in the online path.

PR #756, opened 2024-06-18
fix: Fee sampling comparison metrics when both undefined or buy/sellFeeBps missing

xrsv:
One more minor fix when comparing partially populated fees between the 2 implementations.
Populate with 0 so that we can have correct comparison.

### DIFF
--- a/lib/util/traffic-switch/traffic-switcher-i-token-fee-fetcher.ts
+++ b/lib/util/traffic-switch/traffic-switcher-i-token-fee-fetcher.ts
@@ -2,6 +2,7 @@ import { TrafficSwitcher } from './traffic-switcher'
 import { ITokenFeeFetcher, TokenFeeMap } from '@uniswap/smart-order-router/build/main/providers/token-fee-fetcher'
 import { ProviderConfig } from '@uniswap/smart-order-router/build/main/providers/provider'
 import { log } from '@uniswap/smart-order-router'
+import { BigNumber } from 'ethers'
 
 type Address = string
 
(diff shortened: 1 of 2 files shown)


lib/util/traffic-switch/traffic-switcher-i-token-fee-fetcher.ts
```diff
@@ -33,11 +34,23 @@ export class TrafficSwitcherITokenFeeFetcher extends TrafficSwitcher<ITokenFeeFe
 
     // We have results from both implementations, compare them as a whole.
     // Before comparison, do some cleaning and keep only entries with a fee != 0.
-    // This is needed as different implementations can return empty/null entry, or entry with 0 fees.
+    // This is needed as different implementations can return empty/null entry, or entry with 0 fees (buy/sellFeeBps or both).
     const cleanResult = (result: TokenFeeMap): TokenFeeMap =>
       Object.entries(result)
-        .filter(([_, v]) => !v.buyFeeBps?.eq(0) || !v.sellFeeBps?.eq(0))
-        .reduce((acc, [k, v]) => ({ ...acc, [k]: v }), {})
+        .filter(
```
jsy1218, line 40: good catch...

there's syntactic sugar I think:

```
        .filter(([_, v]) => !v.buyFeeBps?.eq(0) || !v.sellFeeBps?.eq(0))
        .reduce((acc, [k, v]) => ({ ...acc, [k]: {
...v,
buyFeeBps: v.buyFeeBps ?? BigNumber.from(0),
sellFeeBps: v.sellFeeBps ?? BigNumber.from(0),
} }), {})
```

jsy1218 (approved):

mikeki (approved):

PR #766, opened 2024-06-27
fix: extension also hardcode gas on zksync and celo

jsy1218:
after chatting with @ianlapham , we confirmed that extension also needs to have the zksync and celo gas hardcoded

### DIFF
--- a/lib/handlers/quote/quote.ts
+++ b/lib/handlers/quote/quote.ts
@@ -248,6 +248,7 @@ export class QuoteHandler extends APIGLambdaHandler<
 
     const requestSource = requestSourceHeader ?? params.requestQueryParams.source ?? ''
     const isMobileRequest = ['uniswap-ios', 'uniswap-android'].includes(requestSource)
+    const isExtensionRequest = requestSource === 'uniswap-extension'
     const protocols = QuoteHandler.protocolsFromRequest(
       chainId,
       protocolsStr,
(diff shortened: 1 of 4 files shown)


jsy1218:
* **#766** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/766?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

lib/handlers/quote/quote.ts
```diff
@@ -456,12 +457,18 @@ export class QuoteHandler extends APIGLambdaHandler<
       portionAmount: outputPortionAmount, // TODO: name it back to portionAmount
     } = swapRoute
 
-    const estimatedGasUsed = adhocCorrectGasUsed(preProcessedEstimatedGasUsed, chainId, isMobileRequest)
+    const estimatedGasUsed = adhocCorrectGasUsed(
+      preProcessedEstimatedGasUsed,
+      chainId,
+      isMobileRequest,
```
cgkol, line 463: i think this should happen inside the function (pass in the request source instead of parameters for each source)

cgkol (changes requested):

lib/handlers/quote/quote.ts
jsy1218: we do need `isMobileRequest` in that static function https://github.com/Uniswap/routing-api/blob/01dec1424e458cf0236b8167200f9b38ef9f92c1/lib/handlers/quote/quote.ts#L251, because we still have to exclude v2 routing from mobile quote request with version earlier than 1.24.

but I created a generic shouldCorrectGas boolean flag to pass, so that the `adhocCorrectGasUsed` function params shouldn't have to mutate again.

cgkol: @jsy1218 you can pass in the source and do the logic inside the `adhocCorrectGasUsed` and not outside of it because thats where the logic lives right? You can still run `isMobileRequest` but having the source alone should eb good enough for you to make a decision

cgkol (approved):
perfect ✅

jsy1218:
### Merge activity

* **Jun 27, 5:34 PM EDT**: @jsy1218 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/766).

PR #775, opened 2024-07-02
feat: implement quicknode geth vs reth output comparison

jsy1218:
feat: implement quicknode geth vs reth output comparison

feat: implement quicknode geth vs reth output comparison

### DIFF
--- a/lib/rpc/SingleJsonRpcProvider.ts
+++ b/lib/rpc/SingleJsonRpcProvider.ts
@@ -180,7 +180,7 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
     this.logEvaluateLatency()
     this.evaluatingLatency = true
     try {
-      await (this as any)[`${methodName}_EvaluateLatency`](...args)
+      return await (this as any)[`${methodName}_EvaluateLatency`](...args)
     } catch (error: any) {
       this.log.error({ error }, `Encounter error for shadow evaluate latency call: ${JSON.stringify(error)}`)
       // Swallow the error.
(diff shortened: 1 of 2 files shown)


jsy1218:
* **#775** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/775?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

cgkol (approved):
lets do it!

lib/rpc/SingleJsonRpcProvider.ts
```diff
@@ -241,6 +241,22 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
...
+  }
+
+  logRpcResponseMismatch(method: string, otherProvider: SingleJsonRpcProvider) {
+    metric.putMetric(
+      `${this.metricPrefix}_other_provider_${otherProvider.providerId}_method_${method}_rpc_mismatch`,
+      1,
+      MetricLoggerUnit.Count
+    )
+  }
+
```
mikeki, line 259: should these 2 be in `UniJsonRpcProvider`? it kind of feels like `SingleJsonRpcProvider` doesn't need to know about these methods but I could be convinced otherwise

lib/rpc/UniJsonRpcProvider.ts
```diff
@@ -230,7 +231,20 @@ export class UniJsonRpcProvider extends StaticJsonRpcProvider {
         // Within each provider latency shadow evaluation, we should do block I/O,
         // because NodeJS runs in single thread, so it's important to make sure
         // we benchmark the latencies correctly based on the single-threaded sequential evaluation.
-        await provider.evaluateLatency(methodName, args)
+        const evaluatedProviderResponse = await (provider as any)[`evaluateLatency`](methodName, ...args)
```
mikeki, line 234: why do we need to cast to any?

mikeki (commented):
Early comments/questions

xrsv:
@jsy1218 just curious what are the benefits switching to RETH? is it supposed to be faster?

jsy1218:
> @jsy1218 just curious what are the benefits switching to RETH? is it supposed to be faster?

@xrsv Short-term, we benefit from a faster quote (presumably RETH output has the same correctness as GETH). Long-term, the entire ethereum ecosystem benefit from a higher throughput, better performance for full/archive node syncing, better client diversity from GETH, etc. I see Uniswap switching from GETH to RETH for high up-time full-node use case will bring positive effect to those long-term benefits overtime.

lib/rpc/SingleJsonRpcProvider.ts
jsy1218: Eventually all those stuffs should be migrated to the standalone RPC service. I'm not too worried about the small gymnastics like this.

lib/rpc/UniJsonRpcProvider.ts
jsy1218: actually we don't need to. I was copy paste https://github.com/Uniswap/routing-api/blob/5bf66088ba97b2a4707c64ca385a3b8cf90d9c0a/lib/rpc/UniJsonRpcProvider.ts#L294

jsy1218:
@mikeki @cgkol I'm going to merge this PR, as I've written the complete unit test coverage. I'm slightly uncertain about the ethers v5.5 style [send](https://github.com/ethers-io/ethers.js/blob/v5.7/packages/providers/src.ts/json-rpc-provider.ts#L504) output comparison, hence I want to ship to prod, and see if there's any unexpected output mismatch, due to any implementation glitch in this PR.

PR #779, opened 2024-07-12
feat: remove mobile check

dannydaniil:
- This PR aims to remove the check for mobile, version < 1.24 to exclude V2 routing
- More context can be found in this [Slack](https://uniswapteam.slack.com/archives/C02GYG8TU12/p1720726497647639) thread

### DIFF
--- a/lib/handlers/quote/quote.ts
+++ b/lib/handlers/quote/quote.ts
@@ -33,7 +33,6 @@ import { MetricsLogger } from 'aws-embedded-metrics'
 import { CurrencyLookup } from '../CurrencyLookup'
 import { SwapOptionsFactory } from './SwapOptionsFactory'
 import { GlobalRpcProviders } from '../../rpc/GlobalRpcProviders'
-import semver from 'semver'
 import { adhocCorrectGasUsed } from '../../util/estimateGasUsed'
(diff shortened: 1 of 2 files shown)


lib/handlers/quote/quote.ts
```diff
@@ -651,14 +651,9 @@ export class QuoteHandler extends APIGLambdaHandler<
...
     requestedProtocols: string[] | string | undefined,
-    isMobileRequest: boolean,
-    appVersion: string | undefined,
     forceCrossProtocol: boolean | undefined
   ): Protocol[] | undefined {
-    // We will exclude V2 if isMobile and the appVersion is not present or is lower than 1.24
-    const semverAppVersion = semver.coerce(appVersion)
-    const fixVersion = semver.coerce('1.24')!
-    const excludeV2 = isMobileRequest && (semverAppVersion === null || semver.lt(semverAppVersion, fixVersion))
+    const excludeV2 = false
```
cgkol, line 656: if this is always to false shall we delete it beneath too?

dannydaniil: We can't delete it - we need the check for it. We would need to completely remove the condition.

I left it on purpose to reflect the mainnet condition.

cgkol (approved):

cgkol (approved):
approve for now to unblock

PR #781, opened 2024-07-15
seperate quicknode geth vs. reth metrics via different provider naming

jsy1218:
We want to differentiate the latency metrics between quicknode geth vs reth. We do so by differentiating the provider name, based on part of the url, which is a substring of the domain name:

PR #704, opened 2024-05-21
fix: manually fail over mainnet traffic to quicknode geth

jsy1218:
Infura has evelated timeout on mainnet. planning to manually failover to quicknode geth endpoint.

PR #701, opened 2024-05-17
feat: remove quicknode geth to see pure reth perf

jsy1218:
It's not obvious to me about the quicknode reth perf, because the quicknode latency is now a mix of reth and geth. I'm removing geth shadow calls.

PR #785, opened 2024-07-18
chore: add RPC requests logging for RETH calls slower than GETH

jsy1218:
I observe that RETH RPC calls are slower than GETH calls in Uniswap routing backend. My observation is based on two different time periods:

1. Between may 20th and may 21th, I had a [PR](https://github.com/Uniswap/routing-api/pull/701/files) that only samples quicknode reth and another [PR](https://github.com/Uniswap/routing-api/pull/704) that fail over from infura to quicknode geth. This means, before May 20th, I had the shadow sampling of mix of geth and reth, meanwhile between may 20th and may 21th, I had the shadow sampling of only reth.

request volume dropped in half on may 20th, because I changed from shadow sampling both quicknode geth and reth to shadow sampling only quicknode reth:

If reth were to run faster than geth, then I expect P50 latency to go down, but instead it went up:

2. Recently I created a [PR](https://github.com/Uniswap/routing-api/pull/781) to split geth vs reth latency metrics.

Sampling traffic volume are comparable between geth and reth, just to make sure the benchmarking is on the same traffic scale:

I can see that P50 is also higher on reth than geth:

I do believe this is due to Uniswap specific RPC call pattern, since we make heavy multicall for quoting. e.g. this tenderly [simulation](https://www.tdly.co/shared/simulation/f53ad31e-02fc-4dc7-977d-6f821a63d747)  captures a heavy multicall with hundreds of routes. As mentioned in https://github.com/Uniswap/routing-api/pull/775, from Paradigm flood testing, we see quicknode reth is faster than quicknode geth, so this makes me believe it's uniswap specific RPC call pattern.

I think it makes sense to log the RPC requests when we observe the evaluated latencies are slower than the current provider latencies. This can help further debug why RETH for Uniswap is slower.

### DIFF
--- a/lib/rpc/SingleJsonRpcProvider.ts
+++ b/lib/rpc/SingleJsonRpcProvider.ts
@@ -183,18 +183,19 @@ export class SingleJsonRpcProvider extends StaticJsonRpcProvider {
     this.logEvaluateLatency()
     this.evaluatingLatency = true
     try {
-      if (methodName === 'call') {
-        const transaction: Deferrable<TransactionRequest> = args[0]
-        const blockTag: BlockTag | Promise<BlockTag> = args[1]
+      switch (methodName) {
+        case CALL_METHOD_NAME:
+          const transaction: Deferrable<TransactionRequest> = args[0]
+          const blockTag: BlockTag | Promise<BlockTag> = args[1]
 
-        return await this.call_EvaluateLatency(transaction, blockTag, latency)
-      } else if (methodName === 'send') {
-        const underlyingMethodName = args[0]
-        const params: Array<any> = args.splice(1)
+          return await this.call_EvaluateLatency(transaction, blockTag, latency)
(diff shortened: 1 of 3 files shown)


jsy1218:
* **#785** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/785?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

jsy1218:
I can see the new logging showed up in my local AWS routing-api:
Screenshot 2024-07-18 at 8 21 40 PM

mikeki (approved):

jsy1218:
### Merge activity

* **Jul 18, 4:11 PM EDT**: @jsy1218 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/785).

PR #806, opened 2024-08-01
chore(logging): add logging for 500 errors

xrsv:
Adding log for 500 errors.


jsy1218 (approved):
I think there's lint error, which needs to be fixed by running 'npm run fix:prettier' locally.

PR #830, opened 2024-09-13
fix: can not apply cdk when in region ap-southeast-1

anthoai97:
I encountered a small issue when trying to run your CDK stack to test the Swap API in region **ap-southeast-1**.
errr

The problem was caused by the AWS::Lambda::Permission resource. I believe the issue was due to the Lambda alias name being too long. I made a slight adjustment, and now it works as expected.

Complete


jsy1218:
very interesting, we will keep this in mind.

Armss9936 (approved):

PR #843, opened 2024-09-23
feat(unirpc): send 10% bnb traffic to unirpc

xrsv:
Preparing the way to start sending traffic to unirpc

- Starting with 10% traffic on BNB chain only.
- Adding `x-uni-service-id` header to track usage
- introduced a `providerNames` field as we need a way to know which provider we use in order to send additional headers (providerUrls are overwritten with actual URL eventually)
- Done ~~Pending: need to add `UNIRPC_0` json secret value with unirpc endpoint [here](https://us-east-2.console.aws.amazon.com/secretsmanager/secret?name=routing-api-rpc-urls-json-primary&region=us-east-2) - pipelines account.~~

Tested on private AWS + E2E tests with mainnet traffic ([dashboard](https://app.datadoghq.com/dashboard/3tp-vmc-q3n/unirpc-service-dashboard?fromUser=true&refresh_mode=paused&tpl_var_serviceid%5B0%5D=routing_api&from_ts=1727128712130&to_ts=1727129781696&live=false))


lib/config/rpcProviderProdConfig.json
```diff
@@ -46,54 +51,61 @@
...
   },
   {
     "chainId": 1,
     "useMultiProviderProb": 1,
     "latencyEvaluationSampleProb": 0.05,
     "healthCheckSampleProb": 0.001,
-    "providerInitialWeights": [1, 0],
-    "providerUrls": ["QUICKNODE_1", "QUICKNODERETH_1"]
+    "providerInitialWeights": [0.01, 0.99],
+    "providerUrls": ["QUICKNODE_1"],
```
jsy1218, line 81: number of weights have to be equal to the number of provider urls

xrsv: good catch - i even had a test on this that wasn't properly setup

PR #856, opened 2024-10-02
fix: native currency routing in routing-api

jsy1218:


jsy1218:
* **#856** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/856?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

cgkol (commented):
lets please addd a test so we dont miss this next time? or if we add something it gets picked up

lib/handlers/schema.ts
```diff
@@ -15,7 +15,7 @@ export type SupportedPoolInRoute = V2PoolInRoute | V3PoolInRoute | V4PoolInRoute
 
 export type V4PoolInRoute = {
   type: 'v4-pool'
-  address: string
+  poolId: string
```
jsy1218, line 18: this was discussed with @dannydaniil , and since pool id is never gonna be a ETH address, we decide to return string. follow up to change upstream routing-api response payload as well.

jsy1218: there;s type issue, i reverted this change.

test/mocha/e2e/quote.test.ts
```diff
@@ -390,6 +390,16 @@ describe('quote', function () {
...
+
+            // if it's exactIn quote, there's a slight chance the first quote request might be cache miss.
+            // but this is okay because each test case retries 3 times, so 2nd exactIn quote is def expected to hit cached routes.
+            // if it's exactOut quote, we should always hit the cached routes.
+            // this is regardless of protocol version.
+            // the reason is because exact in quote always runs before exact out
+            // along with the native or wrapped native pool token address assertions previously
+            // it ensures the cached routes will always cache wrapped native for v2,v3 pool routes
+            // and native for v4 pool routes
+            expect(response.data.hitsCachedRoutes).to.be.true
```
cgkol, line 402: so we expect the first test to always fail? also at this point we always test cahced routed but never the non cached right?

test/mocha/e2e/quote.test.ts
```diff
@@ -2725,9 +3044,49 @@ describe('quote', function () {
...
+                  chain !== ChainId.SEPOLIA &&
+                  pool.tokenIn.address.toLowerCase() === WNATIVE_ON(chain).address.toLowerCase()
+                ) {
+                  nativeOrWrappedNativePoolFound = true
+                }
+              })
+            })
+
+            if (chain === ChainId.SEPOLIA) {
+              expect(nativePoolFound).to.be.true
```
cgkol, line 3076: this should be true not for sepolia but instead for v4 only right?

test/mocha/e2e/quote.test.ts
cgkol: hmm, do we ever test the non cached routes then?

test/mocha/e2e/quote.test.ts
cgkol: yeah definitely, i think this is correct now but soon will be stale, so we gotta keep it real

cgkol: then again, what if these pools are not selected because of liqudiity even though they are there, what would we do?

jsy1218: the test is written that it either requests v4, or requests v2 & v3, so that it must route through v4, as long as v4 is supported on this chain.

jsy1218: the only case then becomes if entire ETH/USDC v4 pool liquidity is withdrawn from the LPers, which is unlikely

cgkol (approved):

test/mocha/e2e/quote.test.ts
jsy1218: offline discussed, we aren't able to, but we will add ticket, and follow-up asap

jsy1218:
### Merge activity

* **Oct 3, 2:12 PM EDT**: @jsy1218 started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/856).

PR #872, opened 2024-10-15
chore: bump sor to 4.7.1 - fix: v4 subgraph provider add ETH base token, and custom fee tier and tick spacings

jsy1218:
release https://github.com/Uniswap/smart-order-router/pull/751

PR #869, opened 2024-10-14
chore: bump sor to 4.7.5 - fix: cached routes cache invalidation

jsy1218:

PR #879, opened 2024-10-18
fix: filter out non-zero hooks address during v4 subgraph cron job

jsy1218:
i just realized a bug, that routing-api/SOR is actually including the non-zero hooks v4 pools, due to how 1) routing-api subgraph cron job 2) SOR get-candidate-pools work together.

for context, yesterday i realized i forgot to pass in some v4 related alpha router params, when i was fixing the cached routes cache invalidation bug https://github.com/Uniswap/routing-api/pull/869/files#r1805582600. then i realized it doesn;t make sense, that it doesn;t make sense that the prod public trading-api endpoint can route through the v4 pool with the hardcoded hook address 0x0000000000000000000000000000000000000020 https://app.warp.dev/block/WQn2uwJFv6SxNzB4kGjOiO, as i was adding this hardcoded hook address in https://github.com/Uniswap/routing-api/pull/872/files#diff-f259fdf1090ba4d745924307d17fcefdeb89bc4a2154cf17ea6f74cdc045eefa. Since [v4PoolsParams](https://github.com/Uniswap/routing-api/pull/872/files#diff-42941f26ebe3855c458055f7f2269f22a2f85265c24cb1c5ac8dd78d20e0a947R486) didnt get passed into [alpha router params](https://github.com/Uniswap/routing-api/pull/869/files#diff-fe7bfddf4e23f6bcb50753b6b4f42dcea2be8ed0339e67401a488827b27f68b2R144), the only possible is that routing-api and SOR can somehow route through the hook v4 pools automatically.

then i realized in in get-candidate-pools, v4 [directswap](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/functions/get-candidate-pools.ts#L537) will just populate whatever v4 subgraph returns, and it goes inside the if condition [if (top2DirectSwapPool.length == 0 && topNDirectSwaps > 0) ](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/functions/get-candidate-pools.ts#L550), if there's no direct swap from subgraph. I was only passing in [emptyV4FeeTickSpacingsHookAddresses](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/functions/get-candidate-pools.ts#L550) inside if that condition as [v4PoolParams](https://github.com/Uniswap/smart-order-router/blob/main/src/routers/alpha-router/functions/get-candidate-pools.ts#L555).

To fix this, we have to ensure during the routing-api cron job time, for any v4 pool, we filter out non-zero hook address pool for now. i was validating on my local by only allowing zero hook address v4 pool on my local routing cron job, and i was able to see v4 pools with 0x0000000000000000000000000000000000000020 hook addresses got filtered out:

```
{
        "id": "0x9bd8e3ed32766adcc1542ef03ea02f99877f4abd8e5b31dd274b3d029eb4483e",
        "token0": {
          "symbol": "ETH",
          "id": "0x0000000000000000000000000000000000000000",
          "derivedETH": "1"
        },
        "token1": {
          "symbol": "USDC",
          "id": "0x1c7d4b196cb0c7b01d743fbc6116a902379c7238",
          "derivedETH": "0.0004345963274718904446220446907986495"
        },
        "feeTier": "500",
        "tick": "-198877",
        "tickSpacing": "10",
        "liquidity": "78419054283340",
        "hooks": "0x0000000000000000000000000000000000000020",
        "totalValueLockedUSD": "4461.45962729090710774877218570905",
        "totalValueLockedETH": "1.938222767770840108752269938125131",
        "totalValueLockedUSDUntracked": "0",
        "sqrtPrice": "3806602437404081577108314"
      },
```

then hitting against my local quote endpoint, it can no longer route through any 0x0000000000000000000000000000000000000020 hook-address v4 pool: https://app.warp.dev/block/5YKy6oKUF2cdHhBFRDEaWs

### DIFF
--- a/lib/cron/cache-pools.ts
+++ b/lib/cron/cache-pools.ts
@@ -5,7 +5,7 @@ import {
   V2SubgraphPool,
   V2SubgraphProvider,
   V3SubgraphPool,
-  V3SubgraphProvider,
+  V3SubgraphProvider, V4SubgraphPool
 } from '@uniswap/smart-order-router'
 import { EventBridgeEvent, ScheduledHandler } from 'aws-lambda'
 import { S3 } from 'aws-sdk'
@@ -23,6 +23,7 @@ import { AWSMetricsLogger } from '../handlers/router-entities/aws-metrics-logger
 import { metricScope } from 'aws-embedded-metrics'
 import * as zlib from 'zlib'
 import dotenv from 'dotenv'
+import { HOOKS_ADDRESSES_ALLOWLIST } from '../util/hooksAddressesAllowlist'
 
 // Needed for local stack dev, not needed for staging or prod
 // But it still doesn't work on the local cdk stack update,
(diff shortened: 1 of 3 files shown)


xrsv (approved):
looks reasonable

jsy1218:
### Merge activity

* **Oct 18, 2:13 PM EDT**: A user merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/879).

PR #881, opened 2024-10-19
VPC endpoint to unirpc

melodylove-uniswap:
1. put routing-api behind a vpc
2. created a vpc endpoint to connect to unirpc vpc endpoint service in uniswap-backend account
3. created route53 to point routing-dev to the vpc endpoint to avoid updating the endpoint on the service side.

### DIFF
--- a/bin/stacks/routing-lambda-stack.ts
+++ b/bin/stacks/routing-lambda-stack.ts
@@ -19,8 +19,8 @@ import * as aws_route53_targets from 'aws-cdk-lib/aws-route53-targets'
 
 const vpcEndpointServiceMap: Record<string, string> = {
   dev: 'com.amazonaws.vpce.us-east-2.vpce-svc-0945550ad67320638',
-  staging: '',
-  prod: '',
+  staging: 'com.amazonaws.vpce.us-east-2.vpce-svc-00e58a4116063039d',
+  prod: 'com.amazonaws.vpce.us-east-2.vpce-svc-04784c683b22cfabb',
 }
 const privateHostedZoneName = 'unihq.org'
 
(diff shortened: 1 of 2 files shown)


bin/stacks/routing-lambda-stack.ts
```diff
@@ -171,6 +244,8 @@ export class RoutingLambdaStack extends cdk.NestedStack {
       runtime: aws_lambda.Runtime.NODEJS_18_X,
       entry: path.join(__dirname, '../../lib/handlers/index.ts'),
       handler: 'quoteHandler',
+      vpc: vpc,
```
jsy1218, line 247: is it possible to set vpc to only caching routing lambda first?

melodylove-uniswap: updated to put both lambda under vpc as we discussed

jsy1218 (approved):

bin/stacks/routing-lambda-stack.ts
```diff
@@ -77,16 +87,79 @@ export class RoutingLambdaStack extends cdk.NestedStack {
...
+          subnetType: ec2.SubnetType.PUBLIC, // Public subnet with internet access
+          cidrMask: 24, // IP range for public subnet
+        },
+        {
+          name: `RoutingPrivateSubnet-${stage}`,
+          subnetType: ec2.SubnetType.PRIVATE_WITH_EGRESS, // Private subnet with access to the internet via NAT Gateway
+          cidrMask: 24, // IP range for private subnet
+        },
+      ],
+      natGateways: 1, // One NAT Gateway for private subnet internet access
```
xrsv, line 107: not sure if related but good to mention here:
on the unirpc side we started hitting a port connections limit once i started ramping up traffic for routing api
It's weird that routing api lambda didn't hit similar issue, as it is doing similar size of http calls to QN directly.
After further digging, I see that routing api default VPC doesn't have a NAT gateway (and can't see active connections dashboards)?
But with this change we are introducing new VPC with NAT gateway, so trying to understand if that might cause issues due to high amount of traffic.
Anyway we could test the new vpc introduction somehow before fully enabling?

xrsv (approved):
thanks - @jsy1218 can help monitor pipeline
as you mentioned, let's block prod deployment in order to test in staging properly

melodylove-uniswap:
added subnets to the vpc to get around the http port issue.

PR #936, opened 2024-12-06
fix(cachedRoutes): add logs for tracing

xrsv:
Adding logs for tracing when we might be using expired cached routes in order for us to better trace and debug those occasions. 
There is a fix we might do to filter those out, but adding logging first will help us debug better the source of the problem.

### DIFF
--- a/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
+++ b/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
@@ -198,34 +198,34 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
     ).length
     const numberOfNotExpiredCachedRoutes = cachedRoutesArr.length - numberOfExpiredCachedRoutes
     if (numberOfExpiredCachedRoutes > 0 && numberOfNotExpiredCachedRoutes > 0) {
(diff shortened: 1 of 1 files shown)


cgkol (approved):
Nice

lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
```diff
@@ -191,6 +191,46 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
...
+      )
+      // and total
+      metric.putMetric(`RoutesDbArrayWithMixedExpiredCachedRoutesTotal`, cachedRoutesArr.length, MetricLoggerUnit.Count)
+      metric.putMetric(
+        `RoutesDbArrayWithMixedExpiredCachedRoutesTotal_${ID_TO_NETWORK_NAME(chainId)}`,
+        cachedRoutesArr.length,
+        MetricLoggerUnit.Count
+      )
+    } else {
+      metric.putMetric(`RoutesDbArrayWithoutMixedExpiredCachedRoutes`, 1, MetricLoggerUnit.Count)
```
jsy1218, line 226: we should prob log with the {optimistic} prefix, because caching routing lambda should have way more expired routes than onlin lambda

xrsv: makes sense - added

jsy1218 (approved):

PR #965, opened 2025-01-15
chore(unirpc): prepare revert base unirpc to 20%

xrsv:
To be checked in only if there are base unirpc timeout issues (reverts unirpc traffic from 100% -> 20%)


jsy1218 (approved):
pre-approve, just in case

Armss9936 (approved):

Armss9936 (approved):

PR #998, opened 2025-02-03
fix(cache): filter out mixed routes containing protocols not requested for caching ONLY

xrsv:
Filter out mixed routes containing protocols not requested

Added integration tests

### DIFF
--- a/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
+++ b/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
@@ -19,6 +19,10 @@ import { CachedRoutesMarshaller } from '../../marshalling/cached-routes-marshall
 import { PromiseResult } from 'aws-sdk/lib/request'
 import { DEFAULT_BLOCKS_TO_LIVE_ROUTES_DB } from '../../../util/defaultBlocksToLiveRoutesDB'
 import { getSymbolOrAddress } from '../../../util/getSymbolOrAddress'
+import { MixedRoute } from '@uniswap/smart-order-router/build/main/routers'
+import { Pair } from '@uniswap/v2-sdk'
+import { Pool as V3Pool } from '@uniswap/v3-sdk'
+import { Pool as V4Pool } from '@uniswap/v4-sdk'
 
 interface ConstructorParams {
   /**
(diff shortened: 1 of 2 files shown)


lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
```diff
@@ -244,7 +248,35 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
         // Using a map to remove duplicates, we will the different percents of different routes.
         // We also filter by protocol, in case we are loading a route from a protocol that wasn't requested
         if (!routesMap.has(routeId) && protocols.includes(cachedRoute.protocol)) {
```
xrsv, line 250: one question for @jsy1218 - I believe it's ok if we return a cachedRout object with empty routes array, correct?
I see this is current behavior as well, but double checking

jsy1218: yes

jsy1218 (approved):

test/mocha/integ/handlers/router-entities/route-caching/dynamo-route-caching-provider.test.ts
```diff
@@ -362,4 +384,82 @@ describe('DynamoRouteCachingProvider', async () => {
...
+      // Route should exist but have no routes since we didn't request V2 protocol
+      expect(route).to.not.be.undefined
+      expect(route?.routes.length).to.equal(0)
+      // Verify the metadata is preserved
+      expect(route?.chainId).to.equal(ChainId.MAINNET)
+      expect(route?.currencyIn.equals(WETH)).to.be.true
+      expect(route?.currencyOut.equals(USDC_MAINNET)).to.be.true
+    })
+
+    it('should not return mixed V2+V3 route if only V2+MIXED are requested', async () => {
```
cgkol, line 419: can you do v4+v2+mixed on the request? also have we confirmed that these tests failed before?

xrsv: - updated tests with V4 in protocols requests as well.
- yes- above tests fail without the fix

xrsv:
closing in favor of a more complete approach: https://github.com/Uniswap/smart-order-router/pull/820

PR #1017, opened 2025-02-14
fix: mixed route eth <-> weth need to remove fake pool

jsy1218:
i forgot to remove the fake pool when we compose the route object in the quote.ts. we cannot return the fake pool in the route object, when we return to trading-api. because trading-api will actually uses the route object to submit to the universal-router-sdk. and universal-router-sdk has no knowledge of the fake pool. ur-sdk has special logic to handle the ETH <-> WETH connections.


jsy1218:
* **#1017** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/1017?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/1017?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

xrsv (approved):

PR #1051, opened 2025-03-10
chore: bump sor to 4.20.4 - log cached routes change metrics

jsy1218:
routing-api side change complimentary of https://github.com/Uniswap/smart-order-router/pull/843/files. Online Routing lambda needs to send the hashed routeId to the async routing lambda, in order for async routing lambda to track how often cached routes changes metric

Tested in routing-api, I can see:

query string param cachedRoutesRouteIds

cachedRoutesNotChanged metrics

### DIFF
--- a/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
+++ b/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
@@ -339,7 +339,12 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
       // if no Item is found it means we need to send a caching request
       if (shouldSendCachingRequest) {
         metric.putMetric('CachingQuoteForRoutesDbRequestSent', 1, MetricLoggerUnit.Count)
-        this.sendAsyncCachingRequest(partitionKey, [Protocol.V2, Protocol.V3, Protocol.V4, Protocol.MIXED], amount, routeId)
+        this.sendAsyncCachingRequest(
+          partitionKey,
+          [Protocol.V2, Protocol.V3, Protocol.V4, Protocol.MIXED],
+          amount,
+          routeId
+        )
         this.setRoutesDbCachingIntentFlag(partitionKey, amount, currentBlockNumber)
       } else {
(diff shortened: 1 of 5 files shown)


lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
```diff
@@ -293,7 +293,7 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
     if (optimistic) {
       // We send an async caching quote
       // we do not await on this function, it's a fire and forget
-      this.maybeSendCachingQuoteForRoutesDb(partitionKey, amount, currentBlockNumber)
+      this.maybeSendCachingQuoteForRoutesDb(partitionKey, amount, currentBlockNumber, first.routes[0].routeId)
```
xrsv, line 296: couple of questions:
- here we use `first.routes[0].routeId` but on line 243 we use `const routeId = routeToString(cachedRoute.route)`. Are those guaranteed to be the same?
- here we only log stats for the `first`route, what about the rest of the routes in `cachedRoutesArr`? (see related [comment](https://github.com/Uniswap/smart-order-router/pull/843/files#r1988048246) in sor PR)

jsy1218: we only persist the first cached route out of the `CachedRoutes` object right?

Although I think we might be able to generalize this by serializing an array of route ids

xrsv: I believe we persist all routes (in case of split)? https://github.com/Uniswap/routing-api/blob/97f43be808f49331d80cc1f27e92a0eae6077e9a/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts#L414

xrsv (approved):

jsy1218:
### Merge activity

* **Mar 11, 6:06 PM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/1051).

PR #1065, opened 2025-04-01
feat: get cached routes to filter out mismatched mixed routes against requested protocol versions

jsy1218:
part 2 is to filter out the cached routes which happens to be mixed route, and when requested protocol versions include MIXED. We need to ensure, when this happens, mixed route needs to have exact requested protocol versions match, e.g. V2,V3, then MIXED routes has to include V2,V3 in protocolsInvolved.

Pending e2e test

### DIFF
--- a/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
+++ b/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
@@ -147,7 +147,7 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
             // requested protocols do not involve MIXED, so there is no need to filter by protocolsInvolved
             || !protocols.includes(Protocol.MIXED)
             // we know MIXED is getting requested, in this case, we need to ensure that protocolsInvolved contains at least one of the requested protocols
-            || protocols.includes(record.protocolsInvolved))
+            || protocols.includes(record.protocolsInvolved.split(',')))
           .sort((a, b) => b.blockNumber - a.blockNumber)
           .slice(0, this.ROUTES_TO_TAKE_FROM_ROUTES_DB)
 
--- a/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
+++ b/lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
@@ -1,4 +1,5 @@
 import {
+  AlphaRouterConfig,
   CachedRoute,
   CachedRoutes,
   CacheMode,
@@ -97,14 +98,15 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
    * @param protocols
    * @protected
    */
-  protected async _getCachedRoute(
+  protected override async _getCachedRoute(
     chainId: ChainId,
     amount: CurrencyAmount<Currency>,
     quoteCurrency: Currency,
     tradeType: TradeType,
     protocols: Protocol[],
     currentBlockNumber: number,
-    optimistic: boolean
+    optimistic: boolean,
+    alphaRouterConfig?: AlphaRouterConfig
   ): Promise<CachedRoutes | undefined> {
     const { currencyIn, currencyOut } = this.determineCurrencyInOut(amount, quoteCurrency, tradeType)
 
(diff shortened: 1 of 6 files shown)


jsy1218:
* **#1065** <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/1065?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/Uniswap/routing-api/1065?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
```diff
@@ -142,6 +142,15 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
...
           // Older routes might not have the protocol field, so we keep them if they don't have it
           .filter((record) => !record.protocol || protocols.includes(record.protocol))
+          // Older routes might not have the protocolsInvolved field, so we keep them if they don't have it
+          .filter(
+            (record) =>
+              !record.protocolsInvolved ||
+              // requested protocols do not involve MIXED, so there is no need to filter by protocolsInvolved
+              !protocols.includes(Protocol.MIXED) ||
+              // we know MIXED is getting requested, in this case, we need to ensure that protocolsInvolved contains at least one of the requested protocols
+              protocols.includes(record.protocolsInvolved.split(','))
```
xrsv, line 152: `protocols.includes(record.protocolsInvolved.split(','))`
shouldn't we here make sure that all `protocolsInvolved` are in `protocols`

jsy1218: yep changed to `protocols.every((protocol) => (record.protocolsInvolved as String).split(',').includes(protocol))`. I also realized that after sending out the PR

test/mocha/e2e/quote.test.ts
```diff
@@ -1495,6 +1495,50 @@ describe('quote', function () {
               }
             })
 
+            // Test that repros https://uniswapteam.slack.com/archives/C08JM8SKMV2/p1742424330903569?thread_ts=1742424322.824189&cid=C08JM8SKMV2
```
xrsv, line 1498: does it make sense to add an integration test here too?
https://github.com/Uniswap/routing-api/blob/a38023ab9611cb1ea768a1ac054f93cd8f507a5c/test/mocha/integ/handlers/router-entities/route-caching/dynamo-route-caching-provider.test.ts

xrsv: I would also run e2e tests against private deployment with this change as well first before merging

lib/handlers/router-entities/route-caching/dynamo-route-caching-provider.ts
```diff
@@ -142,9 +147,42 @@ export class DynamoRouteCachingProvider extends IRouteCachingProvider {
...
           // Older routes might not have the protocol field, so we keep them if they don't have it
           .filter((record) => !record.protocol || protocols.includes(record.protocol))
+          // Older routes might not have the protocolsInvolved field, so we keep them if they don't have it
+          .filter((record) => {
+            return (
+              !record.protocolsInvolved ||
+              !protocols.includes(Protocol.MIXED) ||
+              // We want to roll out the mixed route with UR v1_2 with percent control,
+              // along with the cached routes so that we can test the performance of the mixed route with UR v1_2ss
+              ((alphaRouterConfig?.enableMixedRouteWithUR1_2Percent ?? 0) >= Math.random() * 100 &&
```
xrsv, line 157: as we discussed offline this should be passed as a bool, e.g. `enableMixedRouteWithUR1_2`.
But do we also need to always allow if UR2_0?

jsy1218: yes we should, updated

xrsv (approved):
looks good - if e2e tests + integration tests pass then we should be good

test/mocha/e2e/quote.test.ts
jsy1218: some are flaky,

so first pass is all of them - https://app.warp.dev/block/7SpHBqndNAcr7dTN98i7jf
second pass is on portion tests only - https://app.warp.dev/block/x09MjdRiQK5JMLM3wzX30I

PR #1083, opened 2025-04-25
chore: bump sor to 4.21.2 - feat: skip cachedRoutes for large swaps

xrsv:
Ships https://github.com/Uniswap/smart-order-router/pull/860 (pending checkin)

**Note**: _had to update several e2e tests related to `hitsCachedRoutes`. Also had to introduce a hardhat fork snapshot/revert before/after each test to prevent tests failign due to pool liquidities changes after large swaps (large swaps now are more accurate, discovering more liquidity, altering more pools)_

Added e2e tests

image

image

### DIFF
--- a/lib/graphql/graphql-provider.ts
+++ b/lib/graphql/graphql-provider.ts
@@ -3,9 +3,11 @@ import { ChainId } from '@uniswap/sdk-core'
 import { GraphQLClient, IGraphQLClient } from './graphql-client'
 import {
   GRAPHQL_QUERY_MULTIPLE_TOKEN_INFO_BY_CONTRACTS,
+  GRAPHQL_QUERY_MULTIPLE_TOKEN_PRICE_BY_CONTRACTS,
   GRAPHQL_QUERY_TOKEN_INFO_BY_ADDRESS_CHAIN,
+  GRAPHQL_QUERY_TOKEN_PRICE_BY_ADDRESS_CHAIN,
 } from './graphql-queries'
-import { TokenInfoResponse, TokensInfoResponse } from './graphql-schemas'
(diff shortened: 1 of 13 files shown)


jsy1218 (commented):
think it's fine to not use the feature flag to control the rollout percentage. but we should pass the amount threshold from routing-api, so that we can start with a higher amount.

jsy1218 (approved):

lib/graphql/graphql-provider.ts
```diff
@@ -89,4 +95,20 @@ export class UniGraphQLProvider implements IUniGraphQLProvider {
...
     return this.client.fetchData<TokensInfoResponse>(query, variables)
   }
+
+  async getTokenPrice(chainId: ChainId, address: string): Promise<TokenPriceResponse> {
+    const query = GRAPHQL_QUERY_TOKEN_PRICE_BY_ADDRESS_CHAIN
+    const variables = { chain: this._chainIdToGraphQLChainName(chainId), address: address }
+    return this.client.fetchData<TokenPriceResponse>(query, variables)
+  }
+
+  async getTokensPrice(chainId: ChainId, addresses: string[]): Promise<TokensPriceResponse> {
```
cgkol, line 105: we should get a cache to save calls right? ttl of 5 mins or so?

cgkol: i guess its par of price fetcher right?

cgkol (approved):

lib/graphql/graphql-provider.ts
xrsv: yes there is a 5mins cache in the `TokenPriceProvider` - in the SOR PR [here](https://github.com/Uniswap/smart-order-router/pull/860/files#diff-58ba8f1c16d47dc99153425422da6316add88064cb3e7ba92053db8d97c8dceeR128)

jsy1218 (approved):

PR #1107, opened 2025-05-07
test auto-formatter

yucanliu:
testing the auto formatter for rest of the repo


yucanliu:
I was manually installing the `prettier` which is a different version - and further caused different output. Everything works after a clean repo with `npm install && npm run build`.

PR #1135, opened 2025-06-18
ROUTE-545 allowlist WETH on OPTIMISM

yucanliu:
ROUTE-545 allowlist WETH on OPTIMISM

### DIFF
--- a/lib/cron/cache-pools.ts
+++ b/lib/cron/cache-pools.ts
@@ -312,6 +312,25 @@ const handler: ScheduledHandler = metricScope((metrics) => async (event: EventBr
         } as V4SubgraphPool)
       }
 
+      if (chainId === ChainId.OPTIMISM) {
+        // OPTIMISM ETH/WETH: https://optimistic.etherscan.io/tx/0x5f81f2aa19a50a76a94a30d3d2a9540cb3cd8597c94499a50330e4b6acbef5c1#eventlog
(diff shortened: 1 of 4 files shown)


lib/cron/cache-pools.ts
```diff
@@ -312,6 +312,25 @@ const handler: ScheduledHandler = metricScope((metrics) => async (event: EventBr
...
+          hooks: '0x480dafdb4d6092ef3217595b75784ec54b52e888',
+          liquidity: '173747248900',
+          token0: {
+            id: '0x0000000000000000000000000000000000000000',
+          },
+          token1: {
+            id: '0x4200000000000000000000000000000000000006',
+          },
+          tvlETH: 826,
+          tvlUSD: 1482475,
```
cgkol, line 330: are these TVLs in decimals? seems kinda low right?

cgkol (approved):
🥳

lib/cron/cache-pools.ts
yucanliu: This is extract from a similar pool here: https://app.uniswap.org/explore/pools/optimism/0x1fb3cf6e48F1E7B10213E7b6d87D4c073C7Fdb7b

I dont think the manually increased TVL matter since we simply want to make it non-zero and got considered by the routing algorithm (pool selector)

PR #1145, opened 2025-06-30
fix: add alerts for subgraph cronjob issues

xrsv:
Adding alerts for mainnet/base subgraph cronjob (once validated we can add for rest chains)
We need to be alerted in case our s3 zips stop updating for any reason.

https://linear.app/uniswap/issue/ROUTE-466/create-poolcache-cronjob-alarms


yucanliu (approved):
LGTM - thanks for adding this alert!

PR #1150, opened 2025-07-01
fix: euler hooks factory provider null check

jsy1218:
we are seeing `TypeError: Cannot read properties of undefined (reading 'eulerHooksProvider')` on some subgraph. cron jobs, after https://github.com/Uniswap/routing-api/pull/1143. the reason is because not every subgraph pool cron job has the `eulerHooksFactoryProvider` instantiated, hence we need to have null check protection.

I tested on my personal aws account, and i no longer see erroring out:


xrsv (approved):

jsy1218:
### Merge activity

* **Jul 1, 10:47 PM UTC**: @jsy1218 merged this pull request with [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/1150).

PR #1163, opened 2025-07-16
ROUTE-573, ROUTE-580, ROUTE-591: Allowlisting Bunni, Clanker, and SlippageFeeHook hooks

Ponx:
Added new Bunni hook on BNB
Added new Clanker hooks on Arbitrum and Unichain
Added SlippageFeeHook on Arbitrum


lib/util/v4HooksPoolsFiltering.ts
```diff
@@ -173,13 +173,46 @@ export function v4HooksPoolsFiltering(chainId: ChainId, pools: Array<V4SubgraphP
...
         additionalAllowedPool += 1
       }
 
+      // BUNNI UNICHAIN ETH/weETH
+      if (
+        pool.id.toLowerCase() === '0x6923777072439713c7b8ab34903e0ea96078d7148449bf54f420320d59ede857'.toLowerCase() &&
+        chainId === ChainId.UNICHAIN
+      ) {
+        pool.tvlETH = 1.13 // (https://bunni.xyz/explore/pools/unichain/0x6923777072439713c7b8ab34903e0ea96078d7148449bf54f420320d59ede857)
+        pool.tvlUSD = 4100 // (https://bunni.xyz/explore/pools/unichain/0x6923777072439713c7b8ab34903e0ea96078d7148449bf54f420320d59ede857)
```
yucanliu, line 189: @jsy1218 are these values too small? But I guess it's fine right?

lib/util/hooksAddressesAllowlist.ts
```diff
@@ -5,24 +5,41 @@ import { ADDRESS_ZERO } from '@uniswap/router-sdk'
...
 export const ETH_FLETH_AUTO_WRAP_HOOKS_ADDRESS_ON_BASE = '0x9e433f32bb5481a9ca7dff5b3af74a7ed041a888'
+
 export const BUNNI_HOOKS_ADDRESS_v1_0_ON_L1 = '0x0010d0d5db05933fa0d9f7038d365e1541a41888'
 export const BUNNI_HOOKS_ADDRESS_V1_1_ON_L1 = '0x0000da5dcd7ec49d6ca5554f7b1ca1ce33fa1888'
 export const BUNNI_HOOKS_ADDRESS_V1_1_1_ON_L1 = '0x0000fe59823933ac763611a69c88f91d45f81888'
 export const BUNNI_HOOKS_ADDRESS_v1_0_ON_BASE = '0x0010d0d5db05933fa0d9f7038d365e1541a41888'
 export const BUNNI_HOOKS_ADDRESS_V1_1_ON_BASE = '0x0000da5dcd7ec49d6ca5554f7b1ca1ce33fa1888'
 export const BUNNI_HOOKS_ADDRESS_V1_1_1_ON_BASE = '0x0000fe59823933ac763611a69c88f91d45f81888'
 export const BUNNI_HOOKS_ADDRESS_v1_0_ON_ARBITRUM = '0x0010d0d5db05933fa0d9f7038d365e1541a41888'
+// example pool: 0x75c55eda2c37c47eaf1db8b500171f72f23dc5b16404e904866a6ad1b3a3e537
```
yucanliu, line 16: Is this comment correctly placed?

yucanliu (commented):
LGTM! This is a great start, Alex. Thanks for pushing through the process - do you want to join the routing team as an engineer ;) ? 

Let's wait on @jsy1218 to confirm the TVL amount.

Ponx:
> Another nit comment: let's always add one hook in one PR unless you are 101% sure that all hooks will work. Otherwise, a single hiccup in any one of these hooks would block the whole PR.
> 
> E.g. in this case, the Clanker related hooks don't need a TVL update and could be reviewed and merged much quicker.

Yes, fair point, this should have been several commits, or at least one that separated out the more complex Bunni ones from the simpler Clanker and Slippage Fee ones

lib/util/v4HooksPoolsFiltering.ts
jsy1218: if they are this small, we should just keep as is. We cannot help hook inflate TVL beyond what's being LP'ed.

jsy1218 (approved):

PR #1167, opened 2025-07-24
fix: do not allowlist MEVTaxTestInProd

jsy1218:
someone MEVTaxTestInProd is failing the simulated swaps during quoting https://uniswapteam.slack.com/archives/C03JG5E8N4B/p1753390484943619. Do not allowlist.

PR #1177, opened 2025-08-11
Revert "fix: do not allowlist MEVTaxTestInProd"

yucanliu:
Reverts Uniswap/routing-api#1167 to test in prod as we have enabled the cache invalidation


jsy1218:
@saucepoint is the only LPer for  MEVTaxTestInProd, and its possible the LP position goes out of range again. dont see why we should re-enable ths hook

yucanliu:
seems like this is not needed

PR #1196, opened 2025-08-20
fix: all hooks addresses need to be lower case

jsy1218:
all the hardcoded hooks addresses need to be lower case


jsy1218:
### Merge activity

* **Aug 20, 6:46 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/Uniswap/routing-api/1196).

PR #1202, opened 2025-08-28
chore(alerts): add metric + alerts for growing cached pools zip size

xrsv:
https://linear.app/uniswap/issue/ROUTE-631/alert-when-any-zip-file-goes-above-40mb


cgkol (approved):
✅

PR #1348, opened 2026-01-31
Adding upegs system hook to the allowlist

ChainCreators:
Adding upegs system hook to the allowlist

hook source code:
https://github.com/ChainCreators/Upeg/blob/main/upegs_hook/UpegHook.sol

repo:
https://github.com/ChainCreators/Upeg/tree/main

Examples of buying and selling are here:
https://app.uniswap.org/explore/pools/ethereum/0x5a58bbed3fb03eaa7a07ff77f3dc32fbfb8b3ba763805c4bcb42de02c63a2b5c


Ponx:
Added to allowlist through separate PR. Verifiable here: https://github.com/Uniswap/uniroute-public/blob/f493113296063e46829437d997af21a96d1a595f/src/lib/poolCaching/util/hooksAddressesAllowlist.ts#L294

PR #1362, opened 2026-02-25
Add Bean Planker Static Fee V2 Hook address to Monad allowlist

Ghoulouis:
We are Bean Exchange. 
Add Planker Static Fee V2 hook to Monad allowlist (A fork from Clanker)

Hook address: 
`0x0da6b6c7e6e658f8f32b619a8e59ba84876e68cc`

Example Pool on Monad:
`poolId: 0xb7ff8826c240031e2911a794d3ef98755d0c1365d14470ad5f4cd59179975c71`


Ponx:
@Ghoulouis Can you add some liquidity to the pool so we can run some test swaps through it?

Ghoulouis:
> @Ghoulouis Can you add some liquidity to the pool so we can run some test swaps through it?

I've created some swap transactions so the pool can exchange on both ends (liquidity is quite limited as this is a beta version).
If you need more, please let us know the minimum amount, and we will fulfill your request.

You can track swap orders here.: https://www.geckoterminal.com/vi/monad/pools/0xb7ff8826c240031e2911a794d3ef98755d0c1365d14470ad5f4cd59179975c71

PR #1410, opened 2026-08-18
fix: remove the 20% slippage cap from the quote schema

codyborn:
## What

Removes `.max(20)` from `slippageTolerance` in the quote schema.

Companion PR in the backend monorepo (uniroute carries the same bound): https://github.com/Uniswap/backend/pull/12251

## Why

This bound arrived in [`bc187beb`](https://github.com/Uniswap/routing-api/commit/bc187beb) (2021-06-29, "New responses") as one line in a bulk pass that converted the quote schema to typed Joi and put defensive limits on every field at once — `tokenIn/Out .max(40)`, `amount .max(20)`, `deadline .max(600)`, a recipient regex. The `deadline` line right below it still carries `// 180 mins, same as interface max`, which is the tell: these mirrored the interface's caps, and the interface bounded slippage at 20% then.

The interface has since moved to a 100% max with a warning at 20%, and the Trading API's copy of the bound was removed in January 2025 with product sign-off ([CX-877](https://linear.app/uniswap/issue/CX-877)) — but only in that one layer, leaving routing as the only place still enforcing it.

Effect today: slippage above 20% returns no quote and the app shows a generic "no quotes available" with no explanation. FOT and thin-liquidity tokens routinely need more headroom. Reported four separate times since 2021, and generating routing alert noise since Feb 2025.

No execution-safety argument supports the specific value. `minimumAmountOut` is `amountOut / (1 + slippage)`, which asymptotes toward zero without reaching it — 100% slippage still floors the trade at half the quote. The ceiling belongs to the client, which already has one and already warns.

## Why `min(0)` stays

The SDK asserts `slippage >= 0` in `minimumAmountOut`, so a negative value would throw downstream rather than degrade. `Joi.number()` already rejects `NaN` and `Infinity` by default — verified:

```
20 accept · 21 accept · 50 accept · 1000 accept
-1 REJECT "must be greater than or equal to 0"
NaN REJECT "must be a number"
±Infinity REJECT "cannot be infinity"
```

So dropping `.max(20)` alone is sufficient here. (The uniroute validator needed an explicit finite/non-negative check instead, since it hand-rolls validation on a proto `double` rather than going through Joi.)

## Rollout note

uniroute and routing-api both serve and both carry this bound. Requests over 20% keep failing on whichever path lands second.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


codyborn:
Closing — routing-api is no longer in the serving path, so this bound is dead code and does not need changing.

The live fix is https://github.com/Uniswap/backend/pull/12251, which removes the same 20% cap from uniroute's quote validator.

Leaving the historical note here for anyone who greps this line later: `.max(20)` originated in bc187beb (2021-06-29) as one line in a bulk pass that put defensive limits on every field of the then-new v1 quote schema, mirroring the interface's cap at the time. It was never a deliberate execution-safety decision.

